PR #77, opened 2018-07-24
Add a dedicated build stage for testing

DonSeannelly:


DonSeannelly:
There seems to be some odd URL formatting going on when certain filters are not selected. For instance, here's the URL when I selected one of each filter type: `/browse?text=&currPage=1&limit=20&length=course&level=k-12&&orderBy=date&sortType=-1` but when it is all cleared I have `/browse?text=&currPage=1&limit=20&&&` with a few trailing & operators.

Perhaps we could strip these off?

PR #91, opened 2018-09-03
File Upload Optimizations

gshaw1997:
Adds support for multipart uploads.
Minor refactoring.


src/drivers/S3Driver.ts
```diff
@@ -9,35 +15,131 @@ const AWS_S3_ACL = 'public-read';
...
+   *
+   * @param {{
+   *     file: MultipartFileUpload;
+   *     finish?: boolean;
+   *     completedPartList?: CompletedPartList;
+   *   }} params
+   * @returns {Promise<MultipartUploadData>}
+   * @memberof S3Driver
+   */
+  public async processMultipart(params: {
```
DonSeannelly: This method is a little complex for someone unfamiliar with the process. A few comments describing the different parts would be helpful.

PR #94, opened 2018-09-10
Flag Large Files

gshaw1997:
Adds functionality to flag files with `packageable: boolean` property. If a file's size is >= 100MB the file's packageable property will be set to `true`, otherwise the property is set to `false`.
Files with `packageable` property set to `true` will be added to the Readme PDF as a link.


package.json
```diff
@@ -36,9 +36,11 @@
     "dotenv": "^6.0.0",
     "express": "^4.16.2",
     "express-jwt": "^5.3.0",
+    "http": "0.0.0",
```
DonSeannelly: Why have we installed this package? Does it have any benefits over node's http module?

It looks like it only has one release, 5 years ago.

src/drivers/MongoDriver.ts
```diff
@@ -1162,6 +1162,30 @@ export class MongoDriver implements DataStore {
...
+        {$project: {
+          'materials.files': {$filter: {
+            input: '$materials.files',
+            as: 'file',
+            cond: {$eq: ['$$file.name', params.fileName]},
+          }},
+        }},
+      ]).toArray();
+
+      // Should never return more than one object.
```
DonSeannelly: Should we add a check to see if more than one file is returned? That way we can throw an error and catch that in Sentry if something weird starts happening. Thoughts?

src/drivers/S3Driver.ts
```diff
@@ -200,4 +202,14 @@ export class S3Driver implements FileManager {
       return Promise.reject(e);
     }
   }
+
+
+  private sanitizeFileName(name: string): string {
```
DonSeannelly: This function should get some JSDoc love :heart:

src/interfaces/Responder.ts
```diff
@@ -1,4 +1,5 @@
 import { Subject } from 'rxjs';
+import { Response } from 'express';
```
DonSeannelly: An interface does not have the context to utilize types from express. This should be achieved through an alternative method, within its architectural bounds.

package.json
nrwinner: I don't even think this is being used.

DonSeannelly: @nrwinner I'm not positive on this, but since we do import from `http` I think this module might actually be used instead of node's included http module. It's a weird situation that I've never run into before.

src/drivers/MongoDriver.ts
gshaw1997: Since only one element is needed in the result set, the `aggregate` function should be replaced in favor of MongoDB's `findOne` function which will only return one result. `findOne` can be used along with projection and the`$elemMatch` operator to achieve the filtering done within the `aggregate` function.

PR #95, opened 2018-09-11
:bug: Fix bug causing search to fail if search text contained special characters

gshaw1997:
Fix bug causing search to fail if search text contained special characters by escaping regex characters.


src/drivers/MongoDriver.ts
```diff
@@ -2082,3 +2083,28 @@ export function isEmail(value: string): boolean {
   }
   return false;
 }
+
+/**
+ * Escapes Regex character in string
```
DonSeannelly: Does this escape all regex characters, or select characters? This comment should be written more explicitly.

Issue #93, opened 2018-09-07
Add POST endpoint for collections


DonSeannelly:
An endpoint should be created that allows authorized users to submit their learning object to a collection. The collection should be provided as a route parameter, and the route should authorize the user through their cookie.

PR #99, opened 2018-09-12
create add to collection route #93

nvisal1:
Create route for appending a learning object id to a specified collection. The collection is specified through a query string. The learning object id is passed as a param.


src/drivers/express/ExpressAuthRouteDriver.ts
```diff
@@ -114,6 +114,21 @@ export class ExpressAuthRouteDriver {
...
+        const learningObjectId = req.params.id;
+        const collection = req.query.collection;
+        try {
+          LearningObjectInteractor.addToCollection(
+            this.dataStore,
+            learningObjectId,
+            collection,
+          );
+          responder.sendOperationSuccess();
+        } catch (e) {
```
DonSeannelly: This should console out the error for logging.

src/drivers/express/ExpressAuthRouteDriver.ts
```diff
@@ -114,6 +114,21 @@ export class ExpressAuthRouteDriver {
         responder.sendOperationError(e);
       }
     });
+    router.post('/collection/learning-objects/:id', async (req, res) => {
```
DonSeannelly: `collection` should use the plural form `collections` since it is a REST resource collection

src/interactors/LearningObjectInteractor.ts
```diff
@@ -1065,6 +1065,18 @@ export class LearningObjectInteractor {
...
   }
 
+  public static async addToCollection(
+    dataStore: DataStore,
+    learningObjectId: string,
+    collection: string,
+  ): Promise<void> {
+    try {
+      dataStore.addToCollection(learningObjectId, collection);
+    } catch (e) {
```
DonSeannelly: We should also log out the error here

src/interactors/LearningObjectInteractor.ts
```diff
@@ -1065,6 +1065,18 @@ export class LearningObjectInteractor {
     }
   }
 
+  public static async addToCollection(
+    dataStore: DataStore,
+    learningObjectId: string,
+    collection: string,
+  ): Promise<void> {
+    try {
+      dataStore.addToCollection(learningObjectId, collection);
```
gshaw1997: This try catch block won't catch errors thrown in datastore unless call is awaited. It will get bubbled up to the point where the promise is resolved or rejected, in this case it will be the route handler.

PR #101, opened 2018-09-13
Route for filtering learning objects by specified collection 

nvisal1:
'collection' can be optionally passed as a query string to filter learning objects


src/drivers/MongoDriver.ts
```diff
@@ -1303,6 +1347,10 @@ export class MongoDriver implements DataStore {
     if (level) {
       query.levels = { $in: level };
     }
+    if (collectionIds) {
+      // Collection is not a property of learning object, we are filtering by id
+      query._id = { $in: collectionIds };
```
nrwinner: Do we need to use _id here?

src/drivers/MongoDriver.ts
```diff
@@ -1111,6 +1112,14 @@ export class MongoDriver implements DataStore {
         username: string;
       }[] = await this.matchUsers(author, text);
 
+      // If filtering by collection, get array of
+      // learning object ids that are associated with
+      // the specified collection.
+      let collectionIds;
+      if (collection) {
+        collectionIds = await this.getCollectionIds(collection);
```
DonSeannelly: We should probably move this operation into the query structure so that Mongo can handle it within the transaction. Reading all of these IDs into memory could have negative performance effects on the system.

I'm not familiar if the find query structure will allow for this complicated of a query (since the learning object collections are stored in their own Mongo collection). We may need to switch over to an aggregation pipeline.

gshaw1997: Agreed, I believe the current searching algorithm needs to be looked at to see where the aggregation pipeline can be used to cut down on transactions used to query multiple collections.

src/drivers/MongoDriver.ts
```diff
@@ -1321,9 +1327,37 @@ export class MongoDriver implements DataStore {
...
 
+  /**
+   * Get all learning object ids from a specified collection
+   *
+   * @private
+   * @param {string} collection
+   * @returns
+   * @memberof MongoDriver
+   */
+  private async getCollectionIds(collection: string): Promise<string[]> {
```
nvisal1: This function is no longer being used. Should we remove it?

PR #110, opened 2018-09-28
Tests/in memory datastore

nvisal1:
Changes in this PR: 
* Create Mock Datastore 
* Create const to store mock objects used in unit testing
* Create Mock S3 Driver 
* Refactor Library Interactor to be a driver that implements an interface
* Create Mock Library Driver


npm-debug.log
```diff
@@ -0,0 +1,24 @@
+0 info it worked if it ends with ok
```
DonSeannelly: This debug log shouldn't be checked into the repo. There's no useful information in here for other developers or historical reference.

src/interactors/LearningObject.interactor.spec.ts
```diff
@@ -1,95 +1,85 @@
...
-    const learningObjectName = 'testing more contributors 2';
-    return LearningObjectInteractor.loadLearningObject(driver, username, learningObjectName).then(val => {
-      expect.fail();
-      done();
-    }).catch((error) => {
-      expect(error).to.be.a('string');
-      done();
-    });
-  });
+  // it('should load learning object', done => {
```
DonSeannelly: Why is all of this commented out?

PR #111, opened 2018-10-01
Abstract publishing to a module & in-memory datastore

DonSeannelly:
The ability for an authorized user to toggle a learning object's published flag has been abstracted to its own module. This module contains each architectural level of the system for the given functionality, and simply plugs into the AuthRouteDriver as middleware.

The tests/in-memory datastore branch was merged into this PR. 

Changes in that branch include:
* Create Mock Datastore
* Create const to store mock objects used in unit testing
* Create Mock S3 Driver
* Refactor Library Interactor to be a driver that implements an interface
* Create Mock Library Driver


src/drivers/LearningObjectSubmission/SubmissionRouteDriver.ts
```diff
@@ -0,0 +1,57 @@
...
+      responder.sendOperationSuccess();
+    } catch (e) {
+      responder.sendOperationError(e);
+    }
+  }
+  const router: Router = Router();
+  router.patch('/publish', publish);
+  router.patch('/unpublish', unpublish);
+  return router;
+}
```
nvisal1: I like the structure of this route driver. It is much easier to read.

src/drivers/LearningObjectSubmission/SubmissionRouteDriver.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      await togglePublished(
+        this.dataStore,
+        username,
+        id,
+        published,
+      );
+      responder.sendOperationSuccess();
+    } catch (e) {
+      responder.sendOperationError(e);
```
nvisal1: Should we put a console.error here?

src/middleware/jwt.config.ts
```diff
@@ -1,6 +1,4 @@
 import * as jwt from 'express-jwt';
-import * as dotenv from 'dotenv';
-dotenv.config();
```
nvisal1: Why are we removing dotenv? We need this to set our secret and issuer.

DonSeannelly: I removed only the redundant calls. We were setting it in multiple files, which was just re-exposing environment variables multiple times on startup. Now the main module (app.ts) is the only one to call it so that it runs before anything else, but not multiple times.

It sets the environment on `process.env` which is exposed throughout the runtime and not on a file by file basis.

DonSeannelly: Also, it's actually not needed in production because the environment is not set through a .env file. We should in the future wrap the statement with a conditional to check the `NODE_ENV`

PR #138, opened 2018-11-05
Refactor to modules

DonSeannelly:
This PR removes the functionality of toggling the published status for a learning object in the admin dashboard. The published flag is currently being phased out in favor of status values that show where a learning object is in the submission process.

This PR also adds tests for checking if the whitelist functionality is working.


src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -78,6 +78,19 @@ export class AdminLearningObjectInteractor {
     }
   }
 
+  public static async loadFullLearningObject(
+    dataStore: DataStore,
+    fileManager: FileManager,
```
gshaw1997: `fileManager` and `library` do not appear to be used. Can we remove these parameters?

src/SingleFileDownload/RouteHandler.ts
```diff
@@ -0,0 +1,69 @@
...
+        learningObjectId: loId,
+      });
+      if (!open) {
+        res.attachment(filename);
+      }
+      // Set mime type only if it is known
+      if (mimeType) res.contentType(mimeType);
+      stream.pipe(res);
+    } catch (e) {
+      if (e.message === 'Invalid Access') {
```
nvisal1: Can we move these error strings to a constant so that they are easier to maintain

src/SingleFileDownload/SingleFileDownloadInteractor.ts
```diff
@@ -0,0 +1,48 @@
+import { DataStore } from '../interfaces/DataStore';
+import { FileManager } from '../interfaces/FileManager';
+import { Readable } from 'stream';
+
+export async function downloadSingleFile(params: {
```
nvisal1: add JSDoc

src/drivers/express/ExpressAdminRouteDriver.ts
```diff
@@ -124,44 +122,16 @@ export class ExpressAdminRouteDriver {
...
-            req.params.username,
-            id,
-            published,
-          );
-          res.sendStatus(200);
-        } catch (e) {
-          console.error(e);
-          res.status(500).send(e);
-        }
+      async (_, res) => {
```
nvisal1: Can you explain the _ instead of req? I have not seen this before.

src/FileManager/FileInteractor.ts
```diff
@@ -0,0 +1,191 @@
...
+      id: params.uploadId,
+    });
+    params.dataStore.deleteMultipartUploadStatus({ id: params.uploadId });
+    await params.fileManager.abortMultipartUpload({
+      path: uploadStatus.path,
+      uploadId: params.uploadId,
+    });
+  } catch (e) {
+    console.error(e);
+    throw `Could not cancel upload`;
```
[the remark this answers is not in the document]
DonSeannelly: I'm not sure where this crept in from, but good catch. Everything about this is very wrong. We should never be throwing a string.

src/Collections/RouteHandler.ts
```diff
@@ -0,0 +1,54 @@
+import {Request, Response, Router} from 'express';
+import * as CollectionInteractor from './CollectionInteractor';
+import {CollectionDataStore} from './CollectionDataStore';
+
+export function initializeCollectionRouter({ router, dataStore }: { router: Router, dataStore: CollectionDataStore}) {
+  /**
+   * Return all collections {name: string, abvName: string, primaryColor: string, hasLogo: boolean}
```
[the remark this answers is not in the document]
DonSeannelly:
```suggestion
   * Gets the metadata for all collections
```

src/FileManager/FileInteractor.ts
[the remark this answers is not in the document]
DonSeannelly:
```suggestion
    throw new Error(`Could not cancel upload`);
```

src/drivers/express/ExpressAdminRouteDriver.ts
DonSeannelly: I picked this up a while back from a different project I was working on. Using an underscore to indicate that a variable is not used is sort of a semi-standard pattern. [This thread on SO](https://stackoverflow.com/questions/32197927/standard-conventions-for-indicating-a-function-argument-is-unused-in-javascript) talks a little about this.

The more I think about it though, I kind of like the idea of just calling it `req` and having my editor notify me that it is unused.

src/Collections/RouteHandler.ts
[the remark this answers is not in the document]
nvisal1: Yes

src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -3,49 +3,131 @@ import { Request, Response, Router } from 'express';
...
+      });
+      res.sendStatus(200);
     } catch (e) {
       console.error(e);
-      res.status(500).send(e);
+
+      let status = 500;
+
+      // if the error was that the object has a duplicate name, send a 409 error code
+      if (updates && updates.name && e.message === LearningObjectError.DUPLICATE_NAME(updates.name)) {
```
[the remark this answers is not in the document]
nvisal1: I didn't have full context when I wrote this. We can store the error strings in the const.

src/drivers/express/ExpressAdminRouteDriver.ts
DonSeannelly: I agree. I think it's a better option.

```suggestion
      async (req, res) => {
```

PR #139, opened 2018-11-05
:sparkles: Add 404 page for files that do not exist

gshaw1997:
Instead of nondescript  "Internal Server Error" message, the server will now respond will simple 404 page with link a to the learning object.

### DIFF
--- a/src/drivers/MongoDriver.ts
+++ b/src/drivers/MongoDriver.ts
@@ -1221,11 +1221,13 @@ export class MongoDriver implements DataStore {
             },
           },
         );
-      const materials = doc.materials;
+      if (doc) {
+        const materials = doc.materials;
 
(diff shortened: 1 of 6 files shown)


src/drivers/MongoDriver.ts
```diff
@@ -1221,11 +1221,13 @@ export class MongoDriver implements DataStore {
...
-      const materials = doc.materials;
+      if (doc) {
+        const materials = doc.materials;
 
-      // Object contains materials property.
-      // Files array within materials will alway contain one element
-      return materials.files[0];
+        // Object contains materials property.
+        // Files array within materials will alway contain one element
+        return materials.files[0];
```
nrwinner: Should there be a check here to make sure that the files array exists?

PR #156, opened 2018-12-18
Abstract Collection Operations to a Feature Module

DonSeannelly:
The collection operations have been moved to their own module that defines all of the routes, the datastore interface, and interactor methods for all collection operations in the service.


src/Collections/CollectionInteractor.ts
```diff
@@ -0,0 +1,45 @@
+import {CollectionDataStore} from './CollectionDataStore';
+
+export async function fetchCollections(dataStore: CollectionDataStore): Promise<any> {
```
nrwinner: Can we get some JSDoc up in here? 🤔

PR #158, opened 2018-12-19
Implement integration tests for the LearningOutcome module

nrwinner:


src/LearningOutcomes/LearningOutcomeRouteHandler.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+})
+
+describe('PATCH /learning-objects/:id/learning-outcomes/:outcomeId', () => {
+  
+  it('should return a status of 200 and a Learning Outcome', (done) => {
+    request.patch('/learning-objects/someObjectId/learning-outcomes/someLearningOutcomeId')
+    .send({ outcome: { bloom: 'bloom', verb: 'verb' } })
+    .expect('Content-Type', /json/)
+    .expect(200, (err, res) => {
+      expect(res.body).toBeDefined();
```
DonSeannelly: As we discussed, we should update this to run some validation logic on the body to ensure that it actually is an outcome that is returned.

src/LearningOutcomes/LearningOutcomeRouteHandler.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const request = supertest(app);
+
+
+describe('POST /learning-objects/:id/learning-outcomes', () => {
+
+  it('should return a status of 200 and the id of the inserted outcome', (done) => {
+    request
+      .post('/learning-objects/someObjectId/learning-outcomes')
+      .send({ outcome: { bloom: 'bloom', verb: 'verb' } })
```
DonSeannelly: How is this a valid outcome? Doesn't it need to be a bloom and verb from the taxonomy?

src/LearningOutcomes/LearningOutcomeRouteHandler.ts
```diff
@@ -87,6 +87,7 @@ export function initialize({
   router
     .route('/learning-objects/:id/learning-outcomes')
     .post(addLearningOutcome);
+
```
DonSeannelly: Looks like some extra whitespace made its way in here

src/LearningOutcomes/LearningOutcomeRouteHandler.spec.ts
nrwinner: This isn't in fact a valid LearningOutcome. Since the validation logic (creating a new instance of LearningOutcome with the data from the request body) is housed inside the datastore function, and we're using a Mock DataStore for these tests, the function I wrote in the mock datastore to emulate that original datastore function doesn't actually perform validation. I was under the impression that was out of the scope of these tests.

src/LearningOutcomes/LearningOutcomeRouteHandler.ts
nrwinner: This was intentional, I believe it makes the code more readable to space out this statement from the statement below it.
screen shot 2019-02-01 at 11 57 26 am

src/LearningOutcomes/LearningOutcomeRouteHandler.spec.ts
nrwinner: Originally, I thought this was out of scope of the test cases, since the validation logic is housed in the datastore and we're not accessing the datastore (using a mock datastore instead). After giving it some more thought, I've updated my tests to include cases where bad outcome data is submitted in a POST and PATCH request.

PR #161, opened 2018-12-21
Fix issues with duplicate route and user object access

gshaw1997:
**Issue**
In the refactor-to-modules branch of learning-object-service there were two routes that had identical functionality `/users/:username/learning-objects` and `/learning-objects/summary` which both fetched a user's Learning Objects. 
The new-builder branch of clark-client used the public route to load a user's objects on the dashboard and passed in an `accessUnpublished` value of true to load all of the user's objects. This allowed any visitor to see another user's private objects. When the value of `accessUnpublished` was false, not all of the user's objects would appear on their dashboard.

**Solution**
- Removed authenticated `/learning-objects/summary` route, which collided with `/learning-objects/:id`, in favor of the public `/users/:username/learning-objects` route. 
- Created middlewares to separately validate and decode a token from a cookie or Bearer token, enforce authenticated access, and enforce admin privileges. The `processToken` middleware is used at the top of the router, so that all requests will validate and decode the user token if present. This allows `req.user` to be available for use in all route handlers. 
- Moved authorization logic to interactor.


src/interactors/LearningObjectInteractor.ts
```diff
@@ -39,38 +36,35 @@ export class LearningObjectInteractor {
...
     library: LibraryCommunicator;
     username: string;
+    userToken: UserToken;
     accessUnpublished?: boolean;
     loadChildren?: boolean;
     query?: LearningObjectQuery;
   }): Promise<LearningObject[]> {
     try {
+      // Set accessUnpublished
+      const accessUnpublished =
```
DonSeannelly: I think that this is a little hard to follow as a reader. I wouldn't necessarily say that an if statement would be better here since it is in fact a conditional assignment, but at the very least it could use some more commenting.

src/interactors/LearningObjectInteractor.ts
```diff
@@ -762,6 +753,36 @@ export class LearningObjectInteractor {
       return Promise.reject(`Problem removing child. Error: ${e}`);
     }
   }
+
+  /**
+   * Uses passed authorization function to check if user has access to a resource
+   *
+   * @private
+   * @static
+   * @param {({
```
DonSeannelly: This JSDoc seems a little malformed with the semi-colons inside the objects, and property descriptions on the parameters would be helpful.

I personally find that the following notation `@param {UserToken} params.userToken [YOUR DESCRIPTION HERE]` reads easier and works better with most IDE's autocompletion.

DonSeannelly:
@gshaw1997 good job with this! I think this is really a step in the right direction in how we are handling auth 👍

PR #164, opened 2019-01-07
Update clark-entity to 4.0.0-rc

gshaw1997:


src/LearningObjectSubmission/SubmissionInteractor.ts
```diff
@@ -9,18 +10,14 @@ export async function submitForReview(
...
   try {
     const object = await dataStore.fetchLearningObject(id, true, true);
-    // TODO: learning object validation should be moved to the entity level
-    const errorMessage = LearningObjectInteractor.validateLearningObject(object);
-
-    if (errorMessage) {
-      return Promise.reject(errorMessage);
-    }
-
+    const _ = new SubmittableLearningObject(object);
```
DonSeannelly: What is the reason for assigning the initialized object to a constant like this? I don’t see it used anywhere, so couldn’t we just call the constructor?

gshaw1997: Typescript would not allow just using the constructor here.

PR #172, opened 2019-01-10
Add blooms-taxonomy distrubution to the /learning-objects/stats route 

davidnso:
**Troubleshoot docker issues.** 
**check if the head is detached from master**
git branch 
**if the head is detached from master the run...** 
git checkout master 
git pull 
**Then branch off and continue working** 
If you run into a docker issue when trying to spin up the service try running 
rm -rf package-lock.json 
**If the build crashes because it depends on another service to run properly**
docker-compose -f [dependent-service].dev.yml -f [main-service].dev.yml up --build


src/LearningObjectStats/LearningObjectStatStore.ts
```diff
@@ -5,6 +5,7 @@ import {
 import { Db } from 'mongodb';
 import { COLLECTIONS } from '../drivers/MongoDriver';
 
+const ONION_BLOOMS= 'blooms_outcome_distribution'
```
gshaw1997: To make the name here more descriptive of what the string value I'd suggest changing it to something along the lines of
```suggestion
const BLOOMS_DISTRIBUTION_COLLECTION= 'blooms_outcome_distribution'
```

src/LearningObjectStats/LearningObjectStatStore.ts
```diff
@@ -36,7 +38,25 @@ export class LearningObjectStatStore implements LearningObjectStatDatastore {
         unit: 0,
         course: 0,
       },
+      blooms_distribution: {
+        apply: 0,
+        evaluate: 0, 
+        remember: 0,
+      }
     };
+    for(var i=0; i<bloomsData.length; i++){
```
gshaw1997: Before we iterate over `bloomsData` we should check if it is in fact defined, so we don't run into issues trying to access property `length` of undefined.

src/LearningObjectStats/LearningObjectStatStore.ts
```diff
@@ -36,7 +38,25 @@ export class LearningObjectStatStore implements LearningObjectStatDatastore {
...
+      if(bloomsData[i]._id === 'Apply and Analyze'){
```
gshaw1997: For maintainability, create a `const` to store the string ids of each bloom that can be easily modified and accessed. For instance 
```suggestion
const BLOOMS = {
APPLY: 'Apply and Analyze'
...
}
      if(bloomsData[i]._id === BLOOMS.APPLY){
```

gshaw1997: Also, since we have defined cases here of what we want to match, a switch case statement may be more appropriate and easier to read and maintain.

src/LearningObjectStats/LearningObjectStatStore.ts
```diff
@@ -36,7 +50,32 @@ export class LearningObjectStatStore implements LearningObjectStatDatastore {
...
+          case BLOOMS.EVALUATE:
+            stats.blooms_distribution.evaluate = bloomsData[i].value;
+            break;
+          case BLOOMS.REMEMBER:
+            stats.blooms_distribution.remember = bloomsData[i].value;
+            break;
+          default:
+        }
+      }
+    } else {
```
gshaw1997:
```suggestion
    }
```
Non-error log should be removed

src/LearningObjectStats/LearningObjectStatStore.ts
```diff
@@ -36,7 +50,32 @@ export class LearningObjectStatStore implements LearningObjectStatDatastore {
...
+      console.log('BLOOMS_DATA is undefined. ');
```
gshaw1997:
```suggestion
```
Removing else block from above

src/LearningObjectStats/LearningObjectStatStore.ts
```diff
@@ -36,7 +50,32 @@ export class LearningObjectStatStore implements LearningObjectStatDatastore {
...
+    }
```
gshaw1997:
```suggestion
```
Removing else block from above

PR #181, opened 2019-01-16
Recursively update parent object's  timestamps when child object is modified

### LINKED ISSUE #173: Create a mongo query to handle updates when a learning objects' children are changed
This issue goes with the following user story: 
As a user, I want to see the last updated date of a learning object reflect changes to materials and children so that I know the date for any content.

### LINKED ISSUE #175: Update LearningObject deletion to update parents' timestamp if deleted object is a child
When a LearningObject is deleted and is a child of another object, it's parent's timestamp should be updated to the current date.

gshaw1997:
This PR adds logic to learning-object-service to recursively update a child's parent objects' timestamps whenever it is modified.

Closes #175. Closes #173


gshaw1997:
Closes #175. Closed #173

gshaw1997:
This PR is ready to be reviewed. Is labeled WIP because tests are being written.

src/FileManager/FileInteractor.ts
```diff
@@ -43,41 +47,41 @@ export async function startMultipartUpload(params: {
...
-        id: params.uploadId,
-      });
-    } catch (e) {
-      return Promise.reject(e);
-    }
+/**
+ * Processes Multipart Uploads
+ *
+ * @private
+ * @static
```
DonSeannelly: This is not a static member of a class.

```suggestion
```

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -11,6 +11,73 @@ import {
...
+ * }} params
+ * @returns {Promise<void>}
+ */
+export async function updateParentsDate(params: {
+  dataStore: DataStore;
+  childId: string;
+  parentIds?: string[];
+  date: string;
+}): Promise<void> {
+  const parentIds =
```
DonSeannelly: ### Two opinions here

First, I think [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Basic_assignment) the params object would make this more readable:
`const { dataStore, childId, parentIds, date } = params;`

This would allow us to turn this:
```
    await params.dataStore.findParentObjectIds({
      childId: params.childId,
    })
```
Into this:

`await params.dataStore.findParentObjectIds({ childId });`

Second, I think that the way we are leveraging JavaScripts notion of falsyness is decreasing readability of the code (this is a completely personal opinion). I think it might be better to write something like:

```
if(parentIds == null) {
   parentIds = await params.dataStore.findParentObjectIds({
      childId: params.childId,
   })
}
```
Instead of:
```
  const parentIds =
    params.parentIds ||
    (await params.dataStore.findParentObjectIds({
      childId: params.childId,
    }));
```
My reasoning for this is that the above construct transfers between languages, and reads closer to a proper English thought.

I'm open to thoughts on this from all of @Cyber4All/developers

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -11,6 +11,73 @@ import {
...
+      childId: params.childId,
+    }));
+  if (parentIds && parentIds.length) {
+    await Promise.all([
+      // Perform update of all parent dates
+      params.dataStore.updateMultipleLearningObjects({
+        ids: parentIds,
+        updates: { date: params.date },
+      }),
+      // Perform update parents' parents' dates
```
DonSeannelly: Is updating all of the parents of each of the parents (2 hops), or was "parents' parents'" a type-o?

src/LearningObjects/LearningObjectInteractor.ts
nrwinner: In reference to the **first** thought, I think I agree... however destructuring the params object would have the unfortunate side-effect of making function declarations almost twice as long.

In reference to the **second** thought, I agree in this particular scenario, but not as a blanket statement. In this scenario, being wrapper onto two lines is what's really decreasing readability IMHO. I think `const a = c || d` is concise and very readable.

DonSeannelly: @nrwinner you bring up a good point that I missed: there are two ways to do this destructuring. My comment above was referring to **destructuring the params object into local variables** on the first line of the function. However, it is also possible to do the **destructuring inside the function signature**.

I do agree with you, though, that doing the destructuring in the signature does get a little messy when trying to use type annotations.

As per the second point, I agree with you. I think the complexity of calling a function maybe should cause us to avoid it - but it is fine for just working with variables (such as `c` and `d` above)

nrwinner: @DonSeannelly ahh gotcha, in that case I totally agree with your first point.

src/LearningObjects/LearningObjectInteractor.ts
gshaw1997: It updates the parents of each parent.

DonSeannelly: Ah, I understand now. The function takes a LO and updates the date for each parent. If there are parent's it recursively iterates up the chain until it is done. It's not actually two hops because we are currently update the parents of the object in question.

Is that correct?

gshaw1997: Yep, that's correct

nrwinner:
apprové

PR #185, opened 2019-01-17
add status outcome

jbarlow96:
Needed to add status under outcome in order for "change status" to work on the admin dashboard.


src/drivers/MongoDriver.ts
```diff
@@ -1667,6 +1667,8 @@ export class MongoDriver implements DataStore {
         contributors: contributorIds,
         collection: object.collection,
         lock: object.lock,
+        // @ts-ignore
+        status: object.status,
```
DonSeannelly: We should leave a FIXME here to note that after clark-entity is upgraded to include the status property, this tsignore should be removed.

PR #188, opened 2019-01-23
New Access Groups 

nvisal1:
- Remove current admin middleware usage 
- Implement new AuthGaurd that checks for correct accessGroup strings 
- AccessGroup verification now occurs at the interactor level 
- Fix broken togglePublish and deleteLearningObject functions in AdminLearningObjectInteractor
- Reimplement 401 http status error when correct accessGroup is not present


src/drivers/MongoDriver.ts
```diff
@@ -1067,13 +1067,13 @@ export class MongoDriver implements DataStore {
   ): Promise<LearningObject[]> {
     return await Promise.all(
       docs.map(async doc => {
-        const author = await this.fetchUser(doc.authorID);
+        const author = await this.fetchUser(doc.authorID)
```
nrwinner: Why did we remove semicolons here?

src/drivers/MongoDriver.ts
```diff
@@ -1067,13 +1067,13 @@ export class MongoDriver implements DataStore {
...
         const learningObject = await this.generateLearningObject(
           author,
           doc,
           full,
         );
-        learningObject.id = doc._id;
+        learningObject.id = doc._id
```
nrwinner: Why did we remove semicolons here?

src/interactors/AuthGuard.ts
```diff
@@ -0,0 +1,25 @@
...
+  CURATOR = 'curator',
+  REVIEWER = 'reviewer',
+  USER = 'user'
+}
+
+export function verifyAccessGroup(userAccessGroups: string[], requiredAccessGroups: string[]) {
+   
+    let hasAccess = false;
+    userAccessGroups.forEach(userGroup => {
+        requiredAccessGroups.forEach(requiredGroup => {
```
nrwinner: Can we do this more efficiently?

src/drivers/express/ExpressAdminRouteDriver.ts
```diff
@@ -2,6 +2,7 @@ import { LearningObject } from '@cyber4all/clark-entity';
 import { Router } from 'express';
 import { AdminLearningObjectInteractor } from '../../interactors/interactors';
 import { DataStore, FileManager, LibraryCommunicator } from '../../interfaces/interfaces';
+import { access } from 'fs';
```
DonSeannelly: Was this perhaps auto-imported? I don't see it used anywhere.

src/drivers/express/ExpressAdminRouteDriver.ts
```diff
@@ -80,24 +84,30 @@ export class ExpressAdminRouteDriver {
...
             this.dataStore,
+            accessGroups,
             page,
             limit,
           );
         }
         res.status(200).send(learningObjects);
       } catch (e) {
         console.error(e);
+        if (e instanceof Error) {
```
DonSeannelly: I really like that we are adding some better error handling logic here 👍. Is there a way we could ensure that if there is an Error thrown, it is in fact for invalid access? Currently, this will catch all errors and so I could see us getting some weird responses in certain failure situations.

Perhaps we could throw the access error with a particular string as its message: 
```
throw new Error('Invalid Access');
```

So that we could do something like:
```
if (e instanceof Error && e.message === 'Invalid Access')
```

nvisal1: Ok, I like this. I was trying to avoid string comparison, but I see your point. After reviewing your refactor-to-modules PR, I noticed the LearningObjectError const. Maybe we can store all error strings in there?

src/drivers/express/ExpressAdminRouteDriver.ts
nvisal1: Yeah, that's an auto import

src/drivers/express/ExpressAdminRouteDriver.ts
DonSeannelly: Yeah, having a constant that is used when throwing and evaluating the error would definitely be a better way to go with this.

src/LearningObjectStats/LearningObjectStatsInteractor.ts
```diff
@@ -17,12 +17,12 @@ export interface LearningObjectStats {
   };
 }
 export interface LearningObjectStatDatastore {
-  fetchStats(params: { query: any }): Promise<LearningObjectStats>;
+  fetchStats(params: { query: any }): Promise<Partial<LearningObjectStats>>;
```
davidnso: does this change cause any issues with other drivers that may use this function?

src/drivers/MongoDriver.ts
nvisal1: This is fixed now

src/drivers/MongoDriver.ts
nvisal1: This is fixed now

src/interactors/AuthGuard.ts
nvisal1: I removed the nested loop. Now there is one loop that compares the array contents with indexof

src/LearningObjectStats/LearningObjectStatsInteractor.ts
nvisal1: This was part of a merge conflict that @gshaw1997 helped me resolve. This code should be up to date with concrete implementations.

gshaw1997: @nvisal1 Was this branch merged with master? The return type should be `Promise<LearningObjectStats>` now.

src/LearningObjectStats/LearningObjectStatsInteractor.ts
```diff
@@ -17,12 +17,12 @@ export interface LearningObjectStats {
...
 }
 
 export async function getStats(params: {
   dataStore: LearningObjectStatDatastore;
   query: any;
-}): Promise<LearningObjectStats> {
+}): Promise<Partial<LearningObjectStats>> {
```
gshaw1997: @nvisal1 Was this branch merged with master? The return type should be `Promise<LearningObjectStats>` now.

src/drivers/express/ExpressAdminRouteDriver.ts
```diff
@@ -179,7 +211,11 @@ export class ExpressAdminRouteDriver {
           res.sendStatus(200);
         } catch (e) {
           console.error(e);
-          res.status(500).send(e);
+          if (e.includes('Invalid Access')) {
```
gshaw1997: How does the interactor know to throw an 'Invalid Access` error here if no information about the requester is being passed to the `toggleLock` method?

src/drivers/express/ExpressAdminRouteDriver.ts
```diff
@@ -192,25 +228,36 @@ export class ExpressAdminRouteDriver {
...
```
gshaw1997: How does the interactor know to throw an 'Invalid Accesserror here if no information about the requester is being passed to thetoggleLock` method?

src/LearningObjectStats/LearningObjectStatsInteractor.ts
nvisal1: Good catch @davidnso

src/drivers/express/ExpressAdminRouteDriver.ts
nvisal1: I updated the toggleLock method. I looked over it because it will be phased out of the system soon.

src/interactors/AuthGuard.ts
```diff
@@ -0,0 +1,25 @@
...
+  REVIEWER = 'reviewer',
+  USER = 'user'
+}
+
+export function verifyAccessGroup(userAccessGroups: string[], requiredAccessGroups: string[]) {
+   
+    let hasAccess = false;
+    userAccessGroups.forEach(userGroup => {
+        if (requiredAccessGroups.indexOf(userGroup) > -1 ) {
+            hasAccess = true;
```
gshaw1997: We should `break` here since we're done after we find a match.

src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -49,12 +57,15 @@ export class AdminLearningObjectInteractor {
     level: string[],
     standardOutcomeIDs: string[],
     text: string,
+    userAccessGroups: string[],
     orderBy?: string,
     sortType?: number,
     page?: number,
     limit?: number,
-  ): Promise<any> {
+  ): Promise<{ total: number; objects: LearningObject[] } | Error> {
```
gshaw1997: Errors are not actually being returned here. A rejected `Promise` with an error message is being returned instead.

src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -82,11 +97,18 @@ export class AdminLearningObjectInteractor {
     fileManager: FileManager,
     library: LibraryCommunicator,
     learningObjectID: string,
-  ): Promise<any> {
+    userAccessGroups: string[]
+  ): Promise<LearningObject | Error> {
```
gshaw1997: Errors are not actually being returned here. A rejected `Promise` with an error message is being returned instead.

src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -95,25 +117,34 @@ export class AdminLearningObjectInteractor {
     username: string,
     id: string,
     published: boolean,
-  ): Promise<void> {
+    userAccessGroups: string[]
+  ): Promise<void | Error> {
```
gshaw1997: Errors are not actually being returned here. A rejected `Promise` with an error message is being returned instead.

src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -125,18 +156,27 @@ export class AdminLearningObjectInteractor {
     fileManager: FileManager,
     username: string,
     learningObjectName: string,
-  ): Promise<void> {
+    library: LibraryCommunicator,
+    userAccessGroups: string[]
+  ): Promise<void | Error> {
```
gshaw1997: Errors are not actually being returned here. A rejected `Promise` with an error message is being returned instead.

src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -146,19 +186,26 @@ export class AdminLearningObjectInteractor {
     library: LibraryCommunicator,
     username: string,
     learningObjectIDs: string[],
-  ): Promise<void> {
+    userAccessGroups: string[]
+  ): Promise<void  | Error> {
```
gshaw1997: Errors are not actually being returned here. A rejected `Promise` with an error message is being returned instead.

src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -167,16 +214,23 @@ export class AdminLearningObjectInteractor {
     fileManager: FileManager,
     id: string,
     learningObject: LearningObject,
-  ): Promise<void> {
+    userAccessGroups: string[]
+  ): Promise<void | Error> {
```
gshaw1997: Errors are not actually being returned here. A rejected `Promise` with an error message is being returned instead.

nvisal1: I was returning Errors at one point and forgot to revert the return type. Good catch

src/interactors/AdminLearningObjectInteractor.ts
```diff
@@ -30,13 +36,20 @@ export class AdminLearningObjectInteractor {
       );
       return response;
     } catch (e) {
-      return Promise.reject(
-        `Problem fetching all Learning Objects. Error: ${e}`,
-      );
+      if (e instanceof Error) {
```
DonSeannelly: Why are we rejecting strings?

src/interactors/AuthGuard.ts
```diff
@@ -0,0 +1,207 @@
...
+            }
+        } else {
+            if (requiredAccessGroups.indexOf(userGroup) > -1) {
+                hasAccess = true;
+                break;
+            }
+        }
+    }
+
+    if(canBeUser) {
```
DonSeannelly: Wouldn't it be better to place this at the top, so that it would short circuit the rest of the processing?

src/interactors/AuthGuard.ts
```diff
@@ -0,0 +1,207 @@
...
+        return true;
+    }
+
+    if (!hasAccess) {
+        throw new Error(LearningObjectError.INVALID_ACCESS(userAccessGroups, requiredAccessGroups));
```
DonSeannelly: Is this error message meaningful to the caller, since they are the one providing both the user access groups and the required access groups? Why not just have this function return a boolean. Perhaps something like 

```
return hasAccess || canBeUser;
```

Double check my logic, but I think that would allow us to get rid of the last two if-blocks.

src/interactors/AuthGuard.ts
```diff
@@ -0,0 +1,207 @@
...
+): boolean {
+    const collectionName = parseCollectionName(userGroup);
+    if (collectionName === requestedCollection) {
+        return true;
+    }
+    return false;
+}
+
+/**
+   * Takes a string and parses the collection name out of it. The userGroup is assumed to be
```
DonSeannelly: What's the rationale for not checking the assumption that `userGroup` is a curator or reviewer and throwing an error in the case that it is not?

src/interactors/AuthGuard.ts
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+   * Takes a string and parses the role name out of it. The userGroup is assumed to be
+   * curator or reviewer
+   *
+   * @param {string} userGroup the access group that belong to the current user
+   *
+   * @returns {string} the role name 
+   */
+function parseRoleName(userGroup: string): string {
```
nrwinner: Can we combine the parseRoleName and parseCollectionName functions so as to not violate DRY?

src/interactors/AuthorizationManager.ts
```diff
@@ -34,8 +53,19 @@ function hasPrivilegedAccess(user: UserToken, collection: string) {
  * @param user information about the user who has initiated a privileged write operation
  * @param collection the name of the collection
  */
-function checkCollectionWriteAccess(user: UserToken, collection: string): boolean {
-  return user.accessGroups.includes(`lead@${collection}`) || user.accessGroups.includes(`curator@${collection}`);
+async function checkCollectionWriteAccess(params: { user: UserToken, dataStore: DataStore, objectId: string }): Promise<boolean> {
+  const regexp = new RegExp('/^[a-f\d]{24}$/i');
+  let property = '_id';
```
gshaw1997: Could you explain what the `property` variable is doing here?

nvisal1: I allowed the checkCollectionWriteAccess function to accept either a learning object id or name. The regex checks to see which of the two were given to the function. The property variable updates accordingly and gets injected into the peek query.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -91,29 +92,31 @@ export async function updateLearningObject(params: {
...
+            updates,
+          });
+          updates.date = Date.now().toString();
+          await params.dataStore.editLearningObject({
+            id: params.id,
+            updates,
+          });
+      } else {
+        error = new Error('User does not have authorization to perform this action');
+        reportError(error);
```
DonSeannelly: I may have miscommunicated before. In this scenario we should not be reporting an error, as this is within the normal execution context. Our intent here is just to signal back to the user that they aren't allowed to do this.

We wouldn't want this to clutter up our reporting dashboard, as this is perfectly normal and won't be alerting us to any bugs/degraded service.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -91,29 +92,31 @@ export async function updateLearningObject(params: {
...
+        return Promise.reject(error);
+      }
+    } catch (e) {
+      error = new Error(`Problem updating learning object ${params.id}. ${e}`);
```
DonSeannelly: However, in this case we do want to report an error, as we have hit a catch block. The specific error that we want to report is in fact `e`, the error that was thrown to get us to this point. We want to absorb that error into our reporting service, and generate a new error message that we reject back to the consumer of the service. The block should look something like this:

```
} catch(e) {
  reportError(e);
  return Promise.reject(new Error(`Problem updating learning object ${params.id}.'));
}
```

You'll notice that in this scenario, the error that was thrown is not added to what is rejected. We contain it to this layer of the application, report it, and then generate something scoped for our consumer.

src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -115,6 +115,9 @@ export function initializePrivate({
     try {
       const id: string = req.params.id;
       updates = req.body.learningObject;
+      if(updates.id) {
```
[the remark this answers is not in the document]
DonSeannelly: @nvisal1 can you fill me in on the purpose of this? I don't think I have context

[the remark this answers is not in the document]
nvisal1: We do not want to reassign a learning object id. It will trigger an entity error and corrupt the learning object.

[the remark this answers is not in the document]
DonSeannelly: Okay, that makes sense to me. Is there a better spot to put this code? Perhaps the interactor, as that should be doing validation?

PR #199, opened 2019-02-04
Remove use of the download lock and publish flag from the system

### LINKED ISSUE #191: Remove usage of the published flag from the system
Throughout our services, mainly the learning object service, the published flag is used in our queries to to pull objects from the database. This needs to change and query the status of an object based on a users access privileges.

### LINKED ISSUE #192: Add authorization to fetch released and submitted revisions based on user privileges.
Queries for learning objects by authorized users (reviewers, curators, editors, admins) should return both the released object and any submitted revisions.

Requires #199

### LINKED ISSUE #193: Remove download locks from the system
Once an object is released there shouldn't be a download lock on it. The use of the download lock has become obsolete since a learning object that is visible to the public must have downloadable content otherwise it is not a true 'released' object.

davidnso:
This PR removes the usage of the published flag and download lock. Additionally, this PR adds authorization on read operations in replace the usage of the published flag. This allows authorized users to read released and unreleased objects.

Closes #191
Closes #192
Closes #193 

Needs Cyber4All/clark-client#613
Needs Cyber4All/clark-client#617


src/drivers/MongoDriver.ts
```diff
@@ -777,7 +778,7 @@ export class MongoDriver implements DataStore {
       full,
     );
 
-    if (!accessUnpublished && !learningObject.published)
+    if (!accessUnreleased && !learningObject.published)
```
nvisal1: Can you explain why the published flag is still being used here?

davidnso: This flag must've slipped through on this branch, it'll be removed on the next commit to check the learning objects status.

src/drivers/MongoDriver.ts
```diff
@@ -789,23 +790,23 @@ export class MongoDriver implements DataStore {
...
+        query.status = RELEASED;
       }
 
       let objectCursor = await this.db
         .collection(COLLECTIONS.LEARNING_OBJECTS)
         .find<LearningObjectDocument>(query);
       const totalRecords = await objectCursor.count();
       objectCursor = this.applyCursorFilters(objectCursor, { page, limit });
-
+      console.log(query)
```
DonSeannelly: Looks like a log statement was left in here

src/interactors/LearningObjectInteractor.ts
```diff
@@ -32,17 +32,17 @@ export class LearningObjectInteractor {
...
-      // Set accessUnpublished
-      let accessUnpublished = params.accessUnpublished;
+      // Set accessUnreleased
+      let accessUnreleased = params.accessUnreleased;
 
-      // If accessUnpublished is unset, set equal to the result of the hasOwnership function
-      if (accessUnpublished === undefined || accessUnpublished === null) {
-        accessUnpublished = await this.hasOwnership({
+      // If accessUnreleased is unset, set equal to the result of the hasOwnership function
+      if (accessUnreleased === undefined || accessUnreleased === null) {
```
DonSeannelly: This could be simplified by using the double equals expression below, so that type coercion is not used and both null/undefined will register as being a bottom value:

```suggestion
      if (accessUnreleased == null) {
```

src/interactors/LearningObjectInteractor.ts
```diff
@@ -32,17 +32,17 @@ export class LearningObjectInteractor {
...
     userToken: UserToken;
-    accessUnpublished?: boolean;
+    accessUnreleased?: boolean;
     loadChildren?: boolean;
     query?: LearningObjectQuery;
   }): Promise<LearningObject[]> {
     try {
-      // Set accessUnpublished
-      let accessUnpublished = params.accessUnpublished;
+      // Set accessUnreleased
```
DonSeannelly: This comment doesn't seem to explain a lot. Unless there's a need to elaborate further, I think it's probably not needed.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -48,9 +48,6 @@ export async function addLearningObject(
       if (!author.emailVerified) {
```
gshaw1997: `LearningObject`'s publish/unpublish is still being used here.

src/drivers/MongoDriver.ts
```diff
@@ -765,7 +766,7 @@ export class MongoDriver implements DataStore {
   async fetchLearningObject(
     id: string,
     full?: boolean,
-    accessUnpublished?: boolean,
+    accessUnreleased?: boolean,
```
gshaw1997: Since we're querying on `status` instead of `published` I would suggest this value be converted to a non-binary value to support all possible values of `status`.

src/drivers/MongoDriver.ts
```diff
@@ -777,7 +778,7 @@ export class MongoDriver implements DataStore {
       full,
     );
 
-    if (!accessUnpublished && !learningObject.published)
+    if (!accessUnreleased && learningObject.status!==RELEASED)
```
gshaw1997: This logic should be done inside of the interactor when authorizing. A request to the datastore should not be made to fetch a resource a user doesn't have access to unless a part of that resource is needed for authorization.

src/drivers/MongoDriver.ts
```diff
@@ -789,23 +790,23 @@ export class MongoDriver implements DataStore {
    * @returns {Cursor<LearningObjectRecord>[]} cursor of literally all objects
    */
   async fetchAllObjects(
-    accessUnpublished?: boolean,
+    accessUnreleased?: boolean,
```
gshaw1997: Since we're querying on `status` instead of `published` I would suggest this value be converted to a non-binary value to support all possible values of `status`.

src/drivers/MongoDriver.ts
```diff
@@ -854,13 +855,13 @@ export class MongoDriver implements DataStore {
   async fetchMultipleObjects(
     ids: string[],
     full?: boolean,
-    accessUnpublished?: boolean,
+    accessUnreleased?: boolean,
```
gshaw1997: Since we're querying on `status` instead of `published` I would suggest this value be converted to a non-binary value to support all possible values of `status`.

src/drivers/MongoDriver.ts
```diff
@@ -901,7 +902,7 @@ export class MongoDriver implements DataStore {
     level: string[];
     standardOutcomeIDs: string[];
     text: string;
-    accessUnpublished?: boolean;
+    accessUnreleased?: boolean;
```
gshaw1997: Since we're querying on `status` instead of `published` I would suggest this value be converted to a non-binary value to support all possible values of `status`.

src/drivers/MongoDriver.ts
```diff
@@ -1049,7 +1050,7 @@ export class MongoDriver implements DataStore {
    * @memberof MongoDriver
    */
   private buildSearchQuery(
-    accessUnpublished: boolean,
+    accessUnreleased: boolean,
```
gshaw1997: Since we're querying on `status` instead of `published` I would suggest this value be converted to a non-binary value to support all possible values of `status`.

src/drivers/MongoDriver.ts
```diff
@@ -37,6 +37,7 @@ export enum COLLECTIONS {
   LO_COLLECTIONS = 'collections',
   MULTIPART_STATUSES = 'multipart-upload-statuses',
 }
+const RELEASED = 'released';
```
nrwinner: We should use the `Status` enum here instead of declaring a separate variable to maintain a single source-of-truth. `Status` can be used like `LearningObject.Status.RELEASED` in our current configuration.

src/drivers/express/ExpressRouteDriver.ts
```diff
@@ -146,19 +146,19 @@ export class ExpressRouteDriver {
       .route('/learning-objects/:username/:learningObjectName')
       .get(async (req, res) => {
         try {
-          let accessUnpublished = false;
+          let accessUnreleased = false;
           const username = req.params.username;
           const cookie = req.cookies.presence;
           if (cookie) {
```
gshaw1997: This logic should be handled inside of the interactor method.

src/drivers/MongoDriver.ts
davidnso: I'd suggest adding another value that holds all the possible values of status because, this flag seems to be used at the top level to check whether or not the user is able to access released or unreleased material. So if they can access unreleased material, my suggestion is, in that case we should utilize all the other values of status when making a query in the MongoDriver.

src/drivers/MongoDriver.ts
davidnso: I wasn't aware that the different statuses were already enums in the learning object entity, this change will be made on the next commit. Thank you

src/drivers/MongoDriver.ts
davidnso: In another branch the "fetchLearningObject" function points to the released collection and all of those objects have a status of released, so would this line still be necessary? The accessUnreleased check would happen in the interactor then.

src/interactors/LearningObjectInteractor.ts
davidnso: Yea these comments aren't needed

src/LearningObjects/LearningObjectInteractor.ts
gshaw1997: The status should be `unreleased` if the user is not verified.

src/LearningObjectSubmission/SubmissionInteractor.ts
```diff
@@ -57,20 +56,32 @@ export async function cancelSubmission(
  * @returns {void}
  */
 export async function createChangelog(params: {
-  dataStore: DataStore,
-  learningObjectId: string,
-  user: UserToken,
-  changelogText: string,
+  dataStore: DataStore;
```
DonSeannelly: I thought this change was odd, and ended up finding [this](https://github.com/prettier/prettier/issues/1944) about why `;` instead of `,`

src/interactors/AuthorizationManager.ts
```diff
@@ -99,3 +139,54 @@ async function userIsOwner(params: { dataStore: DataStore; user: UserToken; obje
...
 }
 
+/**
+ * Checks if accessGroups contains admin or editor
+ *
+ * @param {string[]} accessGroups
+ * @returns {boolean}
+ */
+export function isAdminOrEditor(accessGroups: string[]): boolean {
+  if (
```
DonSeannelly: This function can be simplified to:

```
return accessGroups.includes(UserRole.ADMIN) || accessGroups.includes(UserRole.EDITOR);
```

src/interactors/LearningObjectInteractor.ts
```diff
@@ -7,19 +7,50 @@ import {
...
   updateObjectLastModifiedDate,
   updateParentsDate,
 } from '../LearningObjects/LearningObjectInteractor';
+import { sanitizeText, sanitizeObject } from '../functions';
 
 // file size is in bytes
 const MAX_PACKAGEABLE_FILE_SIZE = 100000000;
 
+const LearningObjectState = {
+  UNRELEASED: [
```
DonSeannelly: This seems kind of confusing since unreleased/review/released are groups, as well as individual statuses. Not sure on the solution here. Maybe we need to define phases within the state transition diagram?

src/drivers/express/ExpressAdminRouteDriver.ts
```diff
@@ -61,44 +54,15 @@ export class ExpressAdminRouteDriver {
...
-          res.sendStatus(200);
-        } catch (e) {
-          console.error(e);
-          if (e.includes('Invalid Access')) {
-            res.status(401).send(e);
-          } else {
-            res.status(500).send(e);
-          }
-        }
+        // Respond to clients that this functionality is now gone
```
nrwinner: Thank you for this comment :heart:

src/interactors/AuthorizationManager.ts
```diff
@@ -54,9 +79,13 @@ function hasPrivilegedAccess(user: UserToken, dataStore: DataStore, objectId: st
...
  */
-async function checkCollectionWriteAccess(params: { user: UserToken, dataStore: DataStore, objectId: string }): Promise<boolean> {
+async function checkCollectionWriteAccess(params: {
+  user: UserToken;
+  dataStore: DataStore;
+  objectId: string;
+}): Promise<boolean> {
   // Regex checks to see if the given objectId string contains an id or a name
-  const regexp = new RegExp('/^[a-f\d]{24}$/i');
+  const regexp = new RegExp('/^[a-fd]{24}$/i');
```
nvisal1: Why did you remove the \ ?

src/interactors/LearningObject.interactor.spec.ts
```diff
@@ -1,48 +1,42 @@
...
+describe('loadUsersObjectSummaries', () => {
   it('should load learning object summary', done => {
-    return LearningObjectInteractor.loadLearningObjectSummary({
-      dataStore: driver,
+    return LearningObjectInteractor.loadUsersObjectSummaries({
+      dataStore,
       library,
       userToken: MOCK_OBJECTS.USERTOKEN,
-      username: MOCK_OBJECTS.USERNAME,
+      username: MOCK_OBJECTS.USERTOKEN.username,
```
nvisal1: Good change

src/interactors/LearningObjectInteractor.ts
nrwinner: @DonSeannelly Can you elaborate on the source of confusion? What do you mean by "groups"?

src/interactors/LearningObject.interactor.spec.ts
```diff
@@ -1,48 +1,42 @@
...
     })
       .then(val => {
         expect(val).to.be.an('array');
         done();
       })
       .catch(error => {
+        console.log('ERR: ', error);
```
nvisal1: Why are we logging the failure of a unit test?

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -166,26 +163,34 @@ export async function updateLearningObject(params: {
...
+        updates,
+      });
+      updates.date = Date.now().toString();
+      await params.dataStore.editLearningObject({
+        id: params.id,
+        updates,
+      });
+    } else {
+      return Promise.reject(
+        new Error('User does not have authorization to perform this action'),
```
nvisal1: We should add this error to the LearningObjectError object and reference it.

src/LearningObjectSubmission/SubmissionInteractor.ts
```diff
@@ -57,20 +56,32 @@ export async function cancelSubmission(
...
 }): Promise<void> {
   try {
-    const hasAccess = hasLearningObjectWriteAccess(params.user, params.dataStore, params.learningObjectId);
+    const hasAccess = hasLearningObjectWriteAccess(
+      params.user,
+      params.dataStore,
+      params.learningObjectId,
+    );
     if (hasAccess) {
       const authorID = await params.dataStore.findUser(params.user.username);
```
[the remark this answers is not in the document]
gshaw1997: This is out of scope of the PR as no code was changed here, just formatting. I'll go ahead and update to streamline to changes.

src/interactors/AuthorizationManager.ts
gshaw1997: This was removed by Prettier, meaning it wasn't formatted properly. Was the intention to match digits?

src/interactors/LearningObjectInteractor.ts
gshaw1997: What part is confusing? Is it the grouping of state of the naming used?

DonSeannelly: @nrwinner @gshaw1997 For instance, there is an object called `LearningObjectState`, which contains an array called `UNRELEASED`, which contains a value in it called `LearningObject.Status.UNRELEASED`.

I don't think that this should hold up this PR, its more of a discussion point we might want to bring up. If we do need to group certain Learning Object statuses together due to some logical meaning, which this PR seems to suggest, I think we should standardize a name for those groupings. 

Right now, in the way that it is presented here, I think that an array having the same name as one of it's members (albeit a different namespace) could be confusing.

### TL;DR 
If we mention an "unreleased" Learning Object in conversation, do we mean a LO with status `unreleased` or a status of either `unreleased`/`rejected`?

PR #208, opened 2019-02-06
Service returns 500 when a user is not found

DonSeannelly:
This PR adds error handling so that if a user does not exist, the route to load their learning objects returns a 404 instead of a 500.

Additionally, two new types, `ServiceError` and `ResourceError`, were introduced so that we can have a semantic understanding of whether or not an error has been caught within another layer of the application. This allows us to tailor the way we handle the error in any given context, as well as easily filter all outbound responses to ensure internal service information is not leaked.

Several promise rejections in async functions were also converted to throw statements.


src/drivers/express/ExpressRouteDriver.ts
```diff
@@ -188,7 +188,9 @@ export class ExpressRouteDriver {
         );
         res.status(200).send(objects.map(obj => obj.toPlainObject()));
       } catch (e) {
-        console.error(e);
+        if (e instanceof Error && e.message === 'User not found') {
```
nrwinner: Should we have an enum of error messages? That way, if we need to update one of the messages, the if-checks don't break throughout the application.

src/drivers/MongoDriver.ts
```diff
@@ -724,7 +727,7 @@ export class MongoDriver implements DataStore {
       .collection(COLLECTIONS.USERS)
       .findOne<UserDocument>(query, { projection: { _id: 1 } });
     if (!userRecord)
-      throw new Error('No user with username or email' + username + ' exists.');
+      throw new Error('No user with username or email ' + username + ' exists.');
```
nrwinner: Literally thank you

src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -88,7 +88,7 @@ export function initializePrivate({
       if (
         object &&
         object.name &&
-        e.message === LearningObjectError.DUPLICATE_NAME(object.name)
+        e.message === `A learning object with name '${object.name}' already exists.`
```
DonSeannelly: I wonder how granular we should go in terms of mapping status codes to the reason for an error. In this case, a ResourceError is thrown with the reason being it was a bad request. However, the intention of this route handler is to respond back with [status code 409](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409), indicating a conflict of state between client and server.

**My question is this**: Should we rely on string comparison (or some other method like this) to pick off situations that we do not have built in to our Error objects and the mapping function that converts them to a HTTP response, or should we have a 1:1 mapping between error reasons and status code responses?

src/Changelogs/ChangelogInteractor.ts
```diff
@@ -29,10 +29,10 @@ export async function createChangelog(params: {
       if (objectId && objectId.length > 0) {
         await params.dataStore.createChangelog(params.learningObjectId, authorID, params.changelogText);
       } else {
-        return Promise.reject(new Error(LearningObjectError.RESOURCE_NOT_FOUND()));
+        return Promise.reject(new ResourceError('Learning Object not found.', ResourceErrorReason.NOT_FOUND));
       }
     } else {
-      return Promise.reject(new Error(LearningObjectError.INVALID_ACCESS()));
+      return Promise.reject(new ResourceError('Invalid Access', ResourceErrorReason.INVALID_ACCESS));
```
DonSeannelly: Should `ResourceErrors` that have an invalid access reason need to be provided with a message? Right now, I'm responding back with the string `'Invalid Access'` each time, but might we want to provide more information here?

src/Changelogs/ChangelogInteractor.ts
```diff
@@ -55,6 +55,7 @@ export async function getRecentChangelog(
       const changelog = await dataStore.fetchRecentChangelog(learningObjectId);
       return changelog;
     } catch (e) {
-      return Promise.reject(e instanceof Error ? e : new Error(LearningObjectError.RESOURCE_NOT_FOUND()));
+      // TODO: Remove falsy resource error
```
gshaw1997: Looks like there is a TODO left here

src/Changelogs/ChangelogDatastore.ts
```diff
@@ -40,7 +40,7 @@ export class ChangelogDataStore {
         );
     } catch (e) {
       reportError(e);
-      return Promise.reject(new Error(LearningObjectError.INTERNAL_ERROR()));
+      return Promise.reject(new ServiceError(ServiceErrorType.INTERNAL));
```
gshaw1997: Is there a need to reject an error? Is there a benefit to rejecting over throwing the error?

src/LearningObjects/LearningObjectRouteHandler.ts
gshaw1997: I prefer the 1:1 mapping between error reasons and status code responses. This allows the message to be adjusted without having to go back and rewrite conditionals. As long as our route driver is doing the mapping I don't see an issue with it mapping reasons to status codes since it uses the HTTP spec.

src/Changelogs/ChangelogInteractor.ts
gshaw1997: I like the option to pass in a message. There may be a reason to provide more information in other cases.

src/errors.ts
```diff
@@ -1,39 +1,46 @@
...
+  INTERNAL = 'InternalServiceError',
+}
+export class ServiceError extends Error {
+  constructor(type: ServiceErrorType) {
+    super('Internal Server Error');
+    this.name = type;
+  }
+}
+
+export enum ResourceErrorReason {
```
nvisal1: We should either use type or reason to describe these enums. Using both is a little confusing.

src/LearningObjects/LearningObjectRouteHandler.ts
nvisal1: I am in favor of having a 1:1 mapping between error reasons and status code responses. I think it will provide us with more consistency. Even though the particular reason may be used in one place at the time of implementation, having a 1:1 mapping will allow future developers to use the error reason elsewhere as needed.

src/Changelogs/ChangelogInteractor.ts
nvisal1: I think 'Invalid Access' should provide context about the access groups required for the requested function.

src/drivers/MongoDriver.ts
```diff
@@ -712,14 +712,27 @@ export class MongoDriver implements DataStore {
    */
   async getUserObjects(username: string): Promise<string[]> {
     try {
-      const authorID = await this.findUser(username);
       const objects = await this.db
-        .collection<{ _id: string }>(COLLECTIONS.LEARNING_OBJECTS)
-        .find({ authorID }, { projection: { _id: 1 } })
+        .collection<{ _id: string }>(COLLECTIONS.USERS)
```
nrwinner: I'm probably just missing something but what is the context surrounding this change here?

DonSeannelly: Originally this function made two different database calls, first to fetch the user's id when given the username, and second to find the Learning Objects for the user. This query now does that in one database transaction.

I made this change as I was working on a solution to the issue at the heart of this PR, not having any notion of resource not found when trying to fetch Learning Objects for a user that doesn't exist. I'm still fixing up a few things with the rest of this, so once that is done this change will have a little more context. For now it just serves as a performance enhancement.

src/Changelogs/ChangelogDatastore.ts
DonSeannelly: There is no advantage in rejecting over throwing since the error is being processed within the main execution context. Both have the same effect in the runtime. There is some speculation that rejecting might be faster on some system architectures, but I haven't seen a study.

I left it as a rejection since @nvisal1 originally wrote it that way when creating the module. I'd be open to creating a style rule if we'd prefer to stick with one over the other.

More info: https://stackoverflow.com/questions/33445415/javascript-promises-reject-vs-throw

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -245,7 +245,8 @@ export async function getLearningObjectChildrenById(
     return await dataStore.loadChildObjects({id: objectId, full: true, status: LearningObjectState.ALL});
   } catch (e) {
     reportError(e);
-    return Promise.reject(new Error(LearningObjectError.RESOURCE_NOT_FOUND()));
+    // TODO: Is this actually a case of it not being found? Can't other faults hit this?
```
nvisal1: This function should just return the datastore function call. No try/catch needed

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -532,6 +533,6 @@ async function checkNameExists(params: {
   });
   // @ts-ignore typescript doesn't think a .id property should exist on the existing object
   if (existing && params.id !== existing.id) {
-    throw new Error(LearningObjectError.DUPLICATE_NAME(params.name));
+    throw new ResourceError(`A learning object with name '${params.name}' already exists.`, ResourceErrorReason.BAD_REQUEST);
```
nvisal1: Can you explain throwing vs rejecting? Why are we throwing the error here?

src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -133,12 +132,12 @@ export function initializePrivate({
       if (
         updates &&
         updates.name &&
-        e.message === LearningObjectError.DUPLICATE_NAME(updates.name)
+        e.message === `A learning object with name '${updates.name}' already exists.`
```
nrwinner: As per our discussion, we shouldn't be using this 409 status code, the response should simply be a 400 to indicate a bad request.

src/errors.ts
```diff
@@ -1,40 +1,72 @@
...
+ * Takes an error object of any kind, and maps it to a status code and message.
+ *
+ * Any error that is not of type ResourceError or ServiceError will default to
+ * an Internal Server Error. If a value is passed in that is not of type Error,
+ * it will be reported and and processed as an Internal Server Error.
+ * @param error the Error to map to HTTP information.
+ * @returns information to convert an Error to a proper HTTP response.
+ */
+export function mapErrorToResponseData(error: Error): { code: number, message: string } {
+  if (!(error instanceof Error)) {
```
nvisal1: Thank you

src/interactors/LearningObjectInteractor.ts
```diff
@@ -73,21 +61,23 @@ export class LearningObjectInteractor {
...
       if (
         !this.hasReadAccess({
-          userToken: params.userToken,
+          userToken,
           resourceVal: params.username,
           authFunction: checkAuthByUsername,
         })
       ) {
-        throw new Error('Invalid access');
+        throw new ResourceError('Invalid Access', ResourceErrorReason.INVALID_ACCESS);
```
nvisal1: throw vs reject?

src/Changelogs/ChangelogDatastore.ts
DonSeannelly: @nvisal1 See my above answer to the question of throw vs. reject

src/LearningObjects/LearningObjectInteractor.ts
DonSeannelly: @gshaw1997 the datastore does not handle any errors itself. It also seems, as per my TODO listed here, that the interactor handles the error improperly. Any error that occurs will result in a resource not found error.

Any errors occurring in this call stack will bubble out to the interactor's try/catch block, where the `mapErrorToResponseData` function will handle creating the correct out-bound response and report to Sentry. 

```
} catch (e) {
      const { code, message } = mapErrorToResponseData(e);
      res.status(code).json({message});
}
```

PR #216, opened 2019-02-12
Add route for children

pzalep1:
Adding route for children. 
This is for the children reordering issue that is open in the client.

Required by Cyber4All/clark-client#579


src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -168,10 +168,25 @@ export function initializePrivate({
...
+  const getLearningObjectChildren = async (req: Request, res: Response) => {
+    try{
+      const id = req.params.id;
+      const children = await LearningObjectInteractor.getLearningObjectChildrenById(
+        dataStore, 
+        id
+      );
+      res.status(200).send(children); 
+    } catch (e) {
+      console.error(e);
```
nvisal1: We should avoid having logs in production. Please remove this.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -206,6 +206,24 @@ export async function getLearningObjectById(
...
+ * @param {string} id the learning object's id
+ */
+export async function getLearningObjectChildrenById(
+  dataStore: DataStore,
+  ObjectId: string,
+){
+  try {
+    return await dataStore.loadChildObjects({id: ObjectId, full: true, accessUnreleased: true});
+  } catch (e) {
+    return Promise.reject(`Problem fetching Children. ${e}`);
```
nvisal1: We want to reject error objects instead of strings. We created an object called `LearningObjectError` that defines the errors that learning object service can throw. Try to implement this object when returning errors. 
```suggestion
    reportError(e);
    return Promise.reject(new Error(LearningObjectError.RESOURCE_NOT_FOUND());
```
The reportError function will hand the error object to Sentry. We then return an expected error to the route handler.

src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -168,10 +168,25 @@ export function initializePrivate({
...
+      res.status(500).send(e);
```
nvisal1: This route should not always throw a 500 when an error occurs. Status code 500 indicates that the server itself is at fault. This may not always be the case. We want to map the error object to an appropriate status code in order to provide the client with more context. Luckily, this function already exists.

```suggestion
      if (e instanceof Error) {
        const status = mapErrorToStatusCode(e);
        res.status(status.code).json({message: status.message});
      } else {
        res.sendStatus(500);
      }
```

src/LearningObjects/LearningObjectRouteHandler.ts
pzalep1: I removed this.

nvisal1:
The build is failing because of Typescript errors. Try importing the mapErrorToStatusCode function and the learningObjectError object in the appropriate files. Let me know if you have any questions.

src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -177,12 +178,7 @@ export function initializePrivate({
       );
       res.status(200).send(children); 
     } catch (e) {
-      if (e instanceof Error) {
-        const status = mapErrorToStatusCode(e);
-        res.status(status.code).json({message: status.message});
-      } else {
-        res.sendStatus(500);
-      }
+      res.status(500).send(e);
```
nvisal1: It looks like this code was reverted by mistake

pzalep1: Changed this with Fix build errors commit

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -206,6 +206,25 @@ export async function getLearningObjectById(
...
+/**
+ * Fetches a learning objects children by ID
+ *
+ * @export
+ * @param {DataStore} dataStore
+ * @param {string} id the learning object's id
+ */
+export async function getLearningObjectChildrenById(
+  dataStore: DataStore,
+  ObjectId: string,
```
gshaw1997: Camel case variable name here

src/drivers/express/ExpressRouteDriver.ts
```diff
@@ -50,7 +50,7 @@ export class ExpressRouteDriver {
         const limit = req.query.limit ? +req.query.limit : null;
 
         let status = req.query.status ? req.query.status : null;
-        if (status && !Array.isArray(status)) {
```
gshaw1997: This should not be removed. Verify that master has been merged into this branch

src/LearningObjects/LearningObjectRouteHandler.ts
```diff
@@ -168,10 +169,29 @@ export function initializePrivate({
...
   };
 
+  const getLearningObjectChildren = async (req: Request, res: Response) => {
+    try{
+      const id = req.params.id;
+      const children = await LearningObjectInteractor.getLearningObjectChildrenById(
+        dataStore, 
+        id
+      );
+      res.status(200).send(children);
```
DonSeannelly: This should use the json method to send children back, so that the outbound encoding is standardized.

```suggestion
      res.status(200).json(children); 
```

src/LearningObjects/LearningObjectInteractor.ts
pzalep1: Resolved with change to camelcase commit.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -206,6 +206,25 @@ export async function getLearningObjectById(
...
 }
 
+/**
+ * Fetches a learning objects children by ID
+ *
+ * @export
+ * @param {DataStore} dataStore
+ * @param {string} id the learning object's id
+ */
+export async function getLearningObjectChildrenById(
```
nvisal1: Can you add a unit test for this new function in `LearningObjectInteractor.spec.ts`?

src/LearningObjects/LearningObjectInteractor.ts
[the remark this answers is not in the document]
pzalep1: Resolved with Change to camelcase commit

src/LearningObjects/LearningObjectInteractor.ts
pzalep1: Added with fa9531312d13c20de04bdab1f3d20b4b71b2bf15.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "learning-object-service",
-  "version": "2.4.2",
+  "version": "2.4.1",
```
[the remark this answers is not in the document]
pzalep1: Resolved with merge with master.

pzalep1:
I have merged master into my branch and it is saying that it is up to date, but it is not because changes in master, I reverted a commit to my branch and that is where the problem lies. I do not know how to un-revert this commit. Sorry this is taking so long.

PR #218, opened 2019-02-12
Convert clark-entity to internal package

### LINKED ISSUE #214: Convert clark-entity to internal package
In efforts to decouple the system from the shared clark-entity library, clark-entity should be converted to an internal package of learning-object service.

The package will be used and maintained inside of this service.

gshaw1997:
This PR removes the usage of the clark-entity package and uses internally declared entities in its place.

Closes #214

### DIFF
--- a/src/LearningObjectSubmission/SubmissionDatastore.ts
+++ b/src/LearningObjectSubmission/SubmissionDatastore.ts
@@ -1,8 +1,8 @@
-import { User, LearningObject } from '@cyber4all/clark-entity';
 import { COLLECTIONS } from '../drivers/MongoDriver';
 import { Db } from 'mongodb';
 import * as ObjectMapper from '../drivers/Mongo/ObjectMapper';
 import { UserDocument } from '../types';
+import { LearningObject, User } from '../entity';
 
 const ERROR_MESSAGE = {
   INVALID_ACCESS: `Invalid access. User must be verified to release Learning Objects`,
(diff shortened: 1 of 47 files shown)


src/LearningObjectSubmission/SubmissionDatastore.ts
```diff
@@ -32,16 +32,6 @@ export class SubmissionDatastore {
         return Promise.reject(ERROR_MESSAGE.INVALID_ACCESS);
       }
 
-      // else
-      const object: { lock: LearningObject.Lock } = await this.db
```
gshaw1997: This was supposed to be removed in #199 but was missed.

DonSeannelly:
All the source code updates here look good. The test suites are having some problems running, though. Once that is fixed this looks good to go 👍

src/LearningObjectSubmission/SubmissionDatastore.ts
```diff
@@ -1,16 +1,16 @@
-import { User, LearningObject } from '@cyber4all/clark-entity';
 import { COLLECTIONS } from '../drivers/MongoDriver';
 import { Db } from 'mongodb';
 import * as ObjectMapper from '../drivers/Mongo/ObjectMapper';
 import { UserDocument } from '../types';
+import { LearningObject, User } from '../entity';
 
 const ERROR_MESSAGE = {
```
davidnso: Are these all the possible errors for a submission? If there are more would it be best to keep the same pattern of extracting errors to a separate file and importing where they are needed? It seems to make testing a bit easier since we can reference those errors in our tests.

gshaw1997: Yes, this is correct. I'm not entirely sure why they are defined at the Datastore level. The error logic related to completing a business use case should also be handled at the interactor level. Theses changes are out of scope of this PR though, so I won't bother updating here. @davidnso Could you open up an issue to regarding this discussion?

PR #229, opened 2019-02-14
Implement route to load user's profile

nrwinner:
This PR begins to address the ongoing problem of having large interactor functions covering multiple business use cases. Specifically, this PR adds an interactor function specifically for retrieving the objects for a users profile based on the requester's access groups and privileges.

### DIFF
--- a/src/interactors/LearningObjectInteractor.ts
+++ b/src/interactors/LearningObjectInteractor.ts
@@ -172,6 +172,18 @@ export class LearningObjectInteractor {
     }
   }
 
+  /**
+   * Retrieve the objects on a user's profile based on the requester's access groups and priveleges
+   *
+   * @static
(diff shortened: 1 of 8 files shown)


src/interactors/LearningObjectInteractor.ts
```diff
@@ -172,6 +172,47 @@ export class LearningObjectInteractor {
...
+   * @static
+   * @param {{
+   *     dataStore: DataStore;
+   *     username: string;
+   *     userToken?: UserToken;
+   *   }} params
+   * @returns {Promise<LearningObject[]>}
+   * @memberof LearningObjectInteractor
+   */
+  public static async loadProfile(params: {
```
nvisal1: Can we get a test for this new function

PR #232, opened 2019-02-15
Fix searching for reviewers/curators

gshaw1997:
This PR fixes searching for reviewers/curators.

**Issue**
When a reviewer or curator search for an object, instead of the result set being filtered out by their search query, the service would return all objects the user had access to. This was due to the search query being overwritten by the condition allowing access to objects by their access groups.

**Solution**
Wrap second set of `$or` conditions in an `$and` operator to avoid overwriting the  `$or` from the initial search query.


src/drivers/MongoDriver.ts
```diff
@@ -164,10 +164,7 @@ export class MongoDriver implements DataStore {
 
     let cursor = this.db
       .collection<LearningObjectDocument>(COLLECTIONS.LEARNING_OBJECTS)
-      .find({
-        ...searchQuery,
-        $or: orConditions,
-      });
+      .find({ ...searchQuery, $and: [{ $or: orConditions }] });
```
gshaw1997: This was the only line changed in the PR. The rest are Prettier changes.

Issue #187, opened 2019-01-22
Implement Learning Object state transitions


davidnso:
A new collection of immutable released learning objects has been created.

Functions are needed to:
- Insert approved objects into the released-objects collection
- Load objects from released-objects collection
- Aggregate data from the released collection and objects collections (holding the working copies) for authorized users.

Needs #192

PR #242, opened 2019-02-22
Add functionality to manage released and working files 

gshaw1997:
This PR adds functions to:
- Retrieve working/released copies of files :heavy_check_mark: 
- Copy working files to released files storage container :heavy_check_mark: 

Needs #187 to implement copying files when an object is released.


gshaw1997:
> Are any of the newly used S3 SDK functions stream based, or are they all simple callbacks that we are just choosing to use promises with?
> 
> The code here looks good, just want to double check we are catching any kinds of errors that might pop up with S3 since it has been [problematic in the past](http://blog.clark.center/2019/01/15/trouble-downstream/).

@DonSeannelly 
The added functions are all callbacks that are converted to promises using their `promise` function. 
I believe this is true for most if not all S3 SDK functions. The difference with the `getObject` method is that we used the `createReadStream` method to get the object's data stream instead of just the meta data that it would return via callback or Promise.

https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getObject-property

PR #256, opened 2019-02-27
Fix Learning Objects not being found if the released copy's name was different from working copy's name

### LINKED ISSUE #250: Editing a released Learning Object's name causes released object to not be found
When an admin/editor edits a released Learning Object's name, the released object is not found.

gshaw1997:
This PR fixes the issue where Learning Objects could not be found if released copy's name was different from working copy's name.

Closes #250

### DIFF
--- a/src/interactors/LearningObjectInteractor.ts
+++ b/src/interactors/LearningObjectInteractor.ts
@@ -331,7 +331,7 @@ export class LearningObjectInteractor {
    * @returns
    * @memberof LearningObjectInteractor
    */
-  private static async loadWorkingObject(params: {
+  private static async loadLearningObjectById(params: {
     dataStore: DataStore;
     learningObjectID: string;
     userToken: UserToken;
--- a/src/interactors/LearningObjectInteractor.ts
+++ b/src/interactors/LearningObjectInteractor.ts
@@ -288,6 +288,89 @@ export class LearningObjectInteractor {
     } catch (e) {
       return Promise.reject(e);
     }
+
+  /**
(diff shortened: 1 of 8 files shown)


src/interactors/LearningObjectInteractor.ts
```diff
@@ -231,22 +231,27 @@ export class LearningObjectInteractor {
 
       const fullChildren = false;
       let loadWorkingCopies = false;
-      const learningObjectID = await dataStore.findLearningObject(
-        username,
-        learningObjectName,
-      );
+
+      const proceedIfPrivileged = (e: Error): null => {
```
DonSeannelly: A comment here describing the use of the error handler would be helpful

DonSeannelly: For instance, why return null over just having a void return type? How is this intended to be used.

src/interactors/LearningObjectInteractor.ts
```diff
@@ -622,58 +803,72 @@ export class LearningObjectInteractor {
...
+      const objectIds = objectRefs.map(obj => obj.id);
+      // Remove objects from library
+      await library.cleanObjectsFromLibraries(objectIds);
+      // Delete objects from datastore
+      await dataStore.deleteMultipleLearningObjects(objectIds);
+      // For each object id
+      objectRefs.forEach(async obj => {
+        // Attempt to delete files
+        const path = `${user.username}/${obj.id}/`;
+        fileManager.deleteAll({ path }).catch(e => {
```
DonSeannelly: We should report this error out to sentry as well so we don't have to check logs manually.

PR #258, opened 2019-02-27
Re-add route to fetch Learning Object's Id & Add authorization logic

gshaw1997:
This PR re-adds the route to fetch a Learning Object's Id (`/learning-objects/:username/:learningObjectName/id`) and refactors authorization logic for this request.

This route is being used by [library-service](https://github.com/Cyber4All/library-service).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "learning-object-service",
-  "version": "2.12.6",
+  "version": "2.12.7",
   "lockfileVersion": 1,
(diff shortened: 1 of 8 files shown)


gshaw1997:
We probably shouldn't be relying on a "ServiceToken" to authorize requests from users. We should instead forward the token of the original requester during service to service communication. Open to comments on this though as I'm not 100% convinced.

This is out of scope of this PR though, so the route was re-added as-is with authorization moved to interactor.

PR #260, opened 2019-02-27
add authorization to functions that modify learning object children

### LINKED ISSUE #255: Authorization logic required for children modification 🚨
No authorization logic is in place for modifying the children of a Learning Object. This should be patched ASAP.

I believe these functions are responsible for children modification and are the ones that need authorization logic added. The functions should verify that the requester has write access to the Learning Object before adding or removing children.

https://github.com/Cyber4All/learning-object-service/blob/6ad4bfa29b9f724adf9f051e2b742be8272d8ce3/src/interactors/LearningObjectInteractor.ts#L983-L1026

davidnso:
This PR implements authorization for adding and removing learning object's children so that only admins, editors, reviewers/curators @ <Parent-Collection> and authors can edit or remove children.

closes #255


src/interactors/LearningObjectInteractor.ts
```diff
@@ -980,20 +981,40 @@ export class LearningObjectInteractor {
...
+      const parentID = await dataStore.findLearningObject(
+        username,
+        parentName,
       );
-      await params.dataStore.setChildren(parentID, params.children);
+      const [status, collection] = await Promise.all([
+        dataStore.fetchLearningObjectStatus(parentID),
+        dataStore.fetchLearningObjectCollection(parentID),
+      ]);
+      this.authorizeReadAccess({
```
gshaw1997: We should check for write access here instead of read access since we are modifying the object.

src/interactors/LearningObjectInteractor.ts
```diff
@@ -1002,20 +1023,38 @@ export class LearningObjectInteractor {
...
+      const parentID = await dataStore.findLearningObject(
+        username,
+        parentName,
       );
-      await params.dataStore.deleteChild(parentID, params.childId);
+      const [status, collection] = await Promise.all([
+        dataStore.fetchLearningObjectStatus(parentID),
+        dataStore.fetchLearningObjectCollection(parentID),
+      ]);
+      this.authorizeReadAccess({ userToken, objectInfo: { author: username, status, collection } })
```
gshaw1997: We should check for write access here instead of read access since we are modifying the object.

src/interactors/LearningObjectInteractor.ts
davidnso: I'll use the 'haslearningobjectwriteaccess' function in the next commit

src/interactors/LearningObjectInteractor.ts
```diff
@@ -980,45 +982,95 @@ export class LearningObjectInteractor {
...
-        id: parentID,
-        date: Date.now().toString(),
-      });
+
+      const hasAccess = await hasLearningObjectWriteAccess(userToken, dataStore, parentID);
+
+      if (hasAccess) {
+        await dataStore.deleteChild(parentID, childId);
+
+        await updateObjectLastModifiedDate({
```
DonSeannelly: I don't have a solution to this, but I feel kind of uneasy about us needing to explicitly call this function to have the timestamp update.

davidnso: The only issue I can think of is a situation where this function isn't added to a block of code that updates a learning object, so the time-stamp is never changed. Can you elaborate?

DonSeannelly: Yeah that's exactly what I'm thinking of too. As time goes on and more code is written, we're bound to run into a place where someone forgets to call the function.

PR #261, opened 2019-02-28
Fix parents route fetching unreleased parents

### LINKED ISSUE #254: Fetch parents route fetches unreleased parents
The fetch parents route needs to be updated to:
- Return only released parents if the user has no privileged access 
- Return parents in the working stage if the user is an admin/editor/curator@collection/reviewer@collection

gshaw1997:
This PR fixes the issue where unreleased parent Learning Objects were returned for unauthorized users when making a request to fetch parents.

Closes #254


src/interactors/LearningObjectInteractor.ts
```diff
@@ -657,19 +661,72 @@ export class LearningObjectInteractor {
...
+          full,
+        });
+      }
+      return [];
     } catch (e) {
-      console.log(e);
-      return Promise.reject(
-        `Problem fetching parent objects for ${params.query.id}. Error: ${e}`,
-      );
+      if (e instanceof ResourceError || e instanceof ServiceError) {
```
DonSeannelly: This seems to be a very common block of code. Maybe we should consider putting this into a function, that way we can simplify all the catch blocks that use it? Probably out of scope for this PR, I just wanted to bring up the idea.

test_environment/globalConfig.json
```diff
@@ -1 +1 @@
-{"mongoUri":"mongodb://127.0.0.1:55321/4775168c-2d44-4d48-b6b7-bc2ff8589740"}
+{"mongoUri":"mongodb://127.0.0.1:42627/17987b01-949b-45b2-b31a-541af53a0650"}
```
DonSeannelly: We should add this to the .gitignore - it's a build artifact of the test runner and doesn't have any meaning between systems.

gshaw1997: This was added to the .gitignore. It needed to be removed from the cache.

PR #267, opened 2019-03-04
Feature/collection stats

nvisal1:
The purpose of this PR is to add a new route for retrieving collection statistics.  

- Extend existing `learning-objects/stats` route to return # of collections

Required by https://github.com/Cyber4All/clark-client/issues/632

Screenshot 
screen shot 2019-03-06 at 7 00 02 am


gshaw1997:
@nvisal1 I thought we were planning on baking this into the existing `/learning-objects/stats` route. Is this not the case anymore? Was there a discussion that I missed?

nvisal1:
> @nvisal1 I thought we were planning on baking this into the existing `/learning-objects/stats` route. Is this not the case anymore? Was there a discussion that I missed?

I wasn't aware of that. We have two separate routes for user stats and learning object stats. Collection stats seem like a logical grouping. Can I get opinions on this @DonSeannelly , @nrwinner , @davidnso

davidnso:
I wasn't aware of that either. It makes more sense to separate the routes based on what is being fetched rather than grouping them under the learning-objects/stats route

DonSeannelly:
> > @nvisal1 I thought we were planning on baking this into the existing `/learning-objects/stats` route. Is this not the case anymore? Was there a discussion that I missed?
> 
> I wasn't aware of that. We have two separate routes for user stats and learning object stats. Collection stats seem like a logical grouping. Can I get opinions on this @DonSeannelly , @nrwinner , @davidnso

Personally, I would view this as a single resource - especially since collections are of Learning Objects which makes them inherently a Learning Object statistic.

The other factor to consider here is that creating multiple routes just adds extra network hops to the client, which wants to just grab the stats and show them.

PR #269, opened 2019-03-04
Refactor interactor level error handling into a function

### LINKED ISSUE #263: Refactor interactor level error handling into a function
This seems to be a very common block of code. Maybe we should consider putting this into a function, that way we can simplify all the catch blocks that use it? Probably out of scope for this PR, I just wanted to bring up the idea.

_Originally posted by @DonSeannelly in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDI2MTIzODM3OQ==/comments/review_comment_

gshaw1997:
This PR refactors the error handling logic on `LearningObjectInteractor.ts` to use the `handleError` function and adds better reporting.

Closes #263

### DIFF
--- a/src/interactors/LearningObjectInteractor.ts
+++ b/src/interactors/LearningObjectInteractor.ts
@@ -1660,15 +1660,14 @@ const bypassNotFoundResourceErrorIfAuthorized = (params: {
 };
 
 /**
- * Handles errors by rejecting Promise if handled, otherwise the error is reported and a ServiceError is thrown
+ * Handles errors by throwing error if handled, otherwise the error is reported and a ServiceError is thrown
  *
- * @template T
- * @param {T} error
- * @returns {Promise<T>}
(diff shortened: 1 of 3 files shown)


src/interactors/LearningObjectInteractor.ts
```diff
@@ -1694,3 +1658,18 @@ const bypassNotFoundResourceErrorIfAuthorized = (params: {
...
 };
+
+/**
+ * Handles errors by rejecting Promise if handled, otherwise the error is reported and a ServiceError is thrown
+ *
+ * @template T
+ * @param {T} error
+ * @returns {Promise<T>}
+ */
+function handleError<T extends Error>(error: T): Promise<T> {
```
DonSeannelly: Do we need to have T extend error, or can we use Error explicitly?

PR #270, opened 2019-03-06
Add method to return child objects in correct order 

pzalep1:
* Added method to return child object in correct order in MongoDriver.ts
* Updated current method to retrieve children in LearningObjectInteractor.ts


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -245,11 +245,23 @@ export async function getLearningObjectChildrenById(
...
+  //fill children array with children learning objects based on order of objects in the database
+  for(let child in childrenIDs){
+    const kid = await dataStore.fetchLearningObject({
+      id: childrenIDs[child],
+      full: true
+    })
+    children.push(kid);
+  } 
+  return children;
 }
```
nvisal1: This logic is correct, but we can reduce it to a single line using the Array.map function

`return childrenIDs.map(async(kid) => await dataStore.fetchLearningObject({
    id: childrenIDs[kid],
    full: true,
  }));`

Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

nvisal1: Also, try installing the TSLint extension in VSCode to avoid CodeFactor issues

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -245,11 +245,23 @@ export async function getLearningObjectChildrenById(
...
+  //Retrieve the ids of the children in the order in which they were set by user 
+  const childrenIDs = await dataStore.findChildObjectIds({
+    parentId: objectId,
+  });  
+
+  //variable to store the full children learning objects 
+  const children: LearningObject[] = [];  
+
+  //fill children array with children learning objects based on order of objects in the database
+  for(let child in childrenIDs){
```
DonSeannelly: I do wonder about the performance hit we might incur here. Perhaps there is a way we can stream children down to the client as they load from the db? Or provide the client with information to make the requests itself? @gshaw1997 @nvisal1 @davidnso Thoughts?

nvisal1:
The build is failing because of a TypeScript error. MockStore implements datastore. Since, you updated the DataStore interface, you need to implement that new function in MockStore.ts. Let me know if you have questions

src/LearningObjects/LearningObjectInteractor.ts
nvisal1: I agree that this implementation could cause a performance hit. However, the purpose of this PR is to return the children in the correct order. I don't see how the client could determine order if it makes several requests.

gshaw1997: I'd suggest making two network requests instead of `n` for the number of children + 1 for the request to get the children's IDs. 

One request to load the child objects for a parent and one to load the children ID's in the correct order. Then performing and in memory sort to return the client the children in the correct order.

nvisal1: I think we can actually just use a single route here. The biggest point of concern is the for loop that iterates over the array of children ids. We can create a new datastore function to fetch learning objects. This new function would accept an array of ids. We can then leverage the $in query operator to fetch all of the objects in a single query. 

docs: https://docs.mongodb.com/manual/reference/operator/query/in/#op._S_in

gshaw1997: @nvisal1 I wasn't suggesting splitting this into separate routes. By requests I meant the requests to the datastore within the interactor. 

The functionality you mentioned above does already exist within the service. That does not solve the problem though. When Mongo queries for the data it adds each result to a set which is an unordered data structure. The only way to get that to be ordered is to order by a property using the `$sort` operator which doesn't work in the case where we want the results ordered by the IDs we sent up. 

The only way to guarantee the result set is in the correct order is to iterate over the array of IDs and result set and sorting in memory. The only other option I see to avoid in memory sorting is to add some sort of `order` property to the document in the DB and use the `$sort` operator to sort by that property. That is not a scalable solution though as many objects can have the same child and may have the child in a different order than other objects.

Performance wise, the cost of an in memory sort on that list is going to be minimal since that list is never likely to become large (likely will never reach ~100 let alone 10^6). The biggest performance concern are the network requests ( the requests to the datastore). From the service perspective, we want to minimize the number of networked requests we have to make to fulfill a user's request.

nvisal1: I understand. Thanks for the clarification.

gshaw1997:
> @nvisal1 I wasn't suggesting splitting this into separate routes. By requests I meant the requests to the datastore within the interactor.
> 
> The functionality you mentioned above does already exist within the service. That does not solve the problem though. When Mongo queries for the data it adds each result to a set which is an unordered data structure. The only way to get that to be ordered is to order by a property using the `$sort` operator which doesn't work in the case where we want the results ordered by the IDs we sent up.
> 
> The only way to guarantee the result set is in the correct order is to iterate over the array of IDs and result set and sorting in memory. The only other option I see to avoid in memory sorting is to add some sort of `order` property to the document in the DB and use the `$sort` operator to sort by that property. That is not a scalable solution though as many objects can have the same child and may have the child in a different order than other objects.
> 
> Performance wise, the cost of an in memory sort on that list is going to be minimal since that list is never likely to become large (likely will never reach ~100 let alone 10^6). The biggest performance concern are the network requests ( the requests to the datastore). From the service perspective, we want to minimize the number of networked requests we have to make to fulfill a user's request.

pzalep1:
> This is a duplicate from the discussions on performance @DonSeannelly brought up. Couldn't figure out a way to quote the discussion from the request changes form.
> 
> I'd suggest making two network requests instead of `n` for the number of children + 1 for the request to get the children's IDs.
> 
> One request to load the child objects for a parent and one to load the children ID's in the correct order. Then performing an in memory sort to return the client the children in the correct order.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -245,11 +245,28 @@ export async function getLearningObjectChildrenById(
...
     id: objectId,
     full: true,
-    status: LearningObjectState.ALL,
+    status: LearningObjectState.ALL
   });
+  //array to return the children in correct order
+  const children: LearningObject[] = [];
+
+  //fill children array with correct order of children
+  for (let child in childrenIDs) {
```
nvisal1: For the in-memory sort, let's use a merge sort (or something similar) instead of nested for loops. 
The worst case for nested for loop implementation is O(n^2). Merge sort worst case is O(n log n).

PR #272, opened 2019-03-06
Add routes to allow uploading to specific user

gshaw1997:
This PR adds routes to allow client to request file upload for specific user.

### DIFF
--- a/src/drivers/express/ExpressAuthRouteDriver.ts
+++ b/src/drivers/express/ExpressAuthRouteDriver.ts
@@ -140,7 +140,13 @@ export class ExpressAuthRouteDriver {
 
     // FILE MANAGEMENT
 
-    // TODO: Deprecate Route
+    /**
+     * @deprecated TODO: Deprecate route
+     * This route should be deprecated once the clients using this route have updated to the new route:
+     * `/users/:username/learning-objects/:learningObjectId/files/:fileId/multipart`
+     * The new route provides more information about the resource being requested and more closely adheres to a RESTful structure
+     *
+     */
     router
       .route('/learning-objects/:id/files/:fileId/multipart')
       .post(async (req, res) => {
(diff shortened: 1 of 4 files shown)


src/drivers/express/ExpressAuthRouteDriver.ts
```diff
@@ -193,6 +195,64 @@ export class ExpressAuthRouteDriver {
...
+        try {
+          const uploadId: string = req.body.uploadId;
+          await FileInteractor.abortMultipartUpload({
+            dataStore: this.dataStore,
+            fileManager: this.fileManager,
+            uploadId,
+          });
+          res.sendStatus(200);
+        } catch (e) {
+          res.status(500).send(e);
```
nvisal1: We should use the `mapErrorToStatus` function instead of sending back 500 errors

src/drivers/express/ExpressAuthRouteDriver.ts
```diff
@@ -139,6 +139,8 @@ export class ExpressAuthRouteDriver {
     );
 
     // FILE MANAGEMENT
+
+    // TODO: Deprecate Route
```
DonSeannelly: Can we explain why this should be deprecated?

src/drivers/express/ExpressAuthRouteDriver.ts
```diff
@@ -193,6 +195,67 @@ export class ExpressAuthRouteDriver {
...
+            fileManager: this.fileManager,
+            uploadId,
+          });
+          res.sendStatus(200);
+        } catch (e) {
+          const response = mapErrorToResponseData(e);
+          res.status(response.code).json(response.message);
+        }
+      });
+    // TODO: Deprecate Route
```
DonSeannelly: Here as well

src/drivers/express/ExpressAuthRouteDriver.ts
```diff
@@ -139,6 +139,14 @@ export class ExpressAuthRouteDriver {
     );
 
     // FILE MANAGEMENT
+
+    /**
+     * @deprecated TODO: Deprecate route
```
nvisal1: 👍

PR #275, opened 2019-03-12
Remove bypass logic to display correct status code

### LINKED ISSUE #274: Display correct error status code on working stage objects
Display correct error status code on working stage objects.  For instance, if an object exists in the working stage but not in the released stage, the API should throw a 401 instead of a 404.  Currently, the route returns a 500 error regardless if the object is not found or the user is unauthorized.

EVogel1999:
Removes bypass logic when retrieving learning object data to display in order to display the correct status code for a given learning object.

Closes #274

### DIFF
--- a/src/interactors/LearningObjectInteractor.ts
+++ b/src/interactors/LearningObjectInteractor.ts
@@ -1676,12 +1676,22 @@ const bypassNotFoundResourceErrorIfAuthorized = (params: {
   error: Error;
   authorizationCases: boolean[];
 }): null | never => {
-  const { error, authorizationCases } = params;
+
+/**
+ * This handler allows execution to proceed if a ResourceError occurs because of a resource not being found.
+ *
+ * @param {Error} error
+ * @returns {null} [Returns null so that the value resolves to null indicating resource was not loaded]
+ */
+const bypassNotFoundResourceError = ({
+  error,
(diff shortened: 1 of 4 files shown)


src/interactors/LearningObjectInteractor.ts
```diff
@@ -238,14 +238,7 @@ export class LearningObjectInteractor {
...
       if (!revision) {
-        const requesterIsAuthor = this.hasReadAccess({
-          userToken,
-          resourceVal: username,
-          authFunction: isAuthorByUsername,
-        }) as boolean;
-        const requesterIsPrivileged =
-          userToken && isPrivilegedUser(userToken.accessGroups);
-        const authorizationCases = [requesterIsAuthor, requesterIsPrivileged];
+        const authorizationCases = [true];
```
DonSeannelly: Can this be explained more? It's not quite clear what this does in isolation.

gshaw1997: This was in order to use the `bypassResourceNotFoundErrorIfAuthorized` handler. By setting the `authorizationCases` to an array of `true`, it allows all users to bypass the error.

To make this less ambiguous, this has been replaced with `bypassResourceNotFoundError` which will bypass for all users.

PR #282, opened 2019-03-15
Fix Permission Issues on Parent Learning Objects

### LINKED ISSUE #280: Broken Authorization When Loading Parents
The route to load parents is responding with all parents, regardless of the fact that some of them may be unreleased and not all users should see them.

DonSeannelly:
## Description
**This is still a work in progress**. I would like to get a review of the scenarios described below before they are fully implemented in code. Thanks!

This PR updates the fetchParentObjects method in the datastore to query the released objects collection when that parameter is specified. Previously, there was an optional property declared in the function signature, but it was not being used inside the function.

## Data Access Conditions
The following scenarios use blue nodes and edges to convey public access, and black nodes/edges to convey privileged access.

### Simple Case: Parents Unreleased and Child Released
In this situation, regular users will only see the released child object as a stand-alone Learning Object. Privileged users (such as editors, admins, reviewers, curators) will be able to see that the Learning Object is included in parent objects which happen to be in the Review Stage.

image

### Child Released, One Parent Released, One Parent Unreleased
This case does not differ much from the above case. All access roles are the same, one of the parents is just now released and can be seen by regular users. Privileged users continue to see both.

image

### Child Released with Revisions, One Parent Released, One Parent Unreleased
Here, a child Learning Object is released, but also has revisions. The relationships mentioned above are maintained, however there is now a node representing the revisions made to the child (labeled as C prime). In this situation, the relationships of the released child transfer to the revised child, but they transfer as uni-directional relationships. This way, the revised child can point to the released parent, but the parent does not directly have a relationship to the child.

image

### Child Released with Revisions, One Parent Released with Revisions, One Parent Unreleased
In this case, we extend the above to include revisions to the parent as well as the child (the parent revisions are denoted as P prime). Under these circumstances, the relationship of C' to P' is **transitive**. C' does not have a direct relationship to P', but instead is linked by its relationship to P. On the other hand, P' does have a uni-directional relationship to C, which then has a bi-directional relationship with C'.

image

## Impact on Users
The way that this plays out through our UI is that revisions of child objects still show the links to their parents. These parents will only be the released objects, or, in the case where the Learning Objects are going through the review stage for the first time, it will be that original copy.

Say we have a reviewer who is looking at a parent object. They will be able to toggle between the released parent and the revised parent. Both of which will point to the released child. They can then navigate to that released child, and from there toggle between the released and revised child object.

## Dependencies
Closes #280

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -22,9 +22,6 @@ const runDevelopmentServer = function() {
   return nodemon({
     script: 'dist/app.js',
     ext: 'js',
-  })
-  .once('exit', () => {
-    process.exit();
   });
 };
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "learning-object-service",
-  "version": "2.14.6",
+  "version": "2.14.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "learning-object-service",
-  "version": "2.14.6",
+  "version": "2.14.7",
   "description": "Learning Object Microservice.",
   "main": "app.ts",
   "scripts": {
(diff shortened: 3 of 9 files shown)


siddharthkaza:
I agree with the thought process for all four cases.

PR #291, opened 2019-03-18
Add the optional hasRevision flag to learning objects

### LINKED ISSUE #277: add hasRevision flag to learning objects that have a duplicate object with a different status
Before we can switch between learning objects that have a revision on Clark. The **hasRevision** flag that is used to check whether or not an object has a duplicate in the objects collection must be passed down so that we can distinguish between objects that have revisions and those that don't. 
From there, privileged users will be able to switch between released and working versions of a learning object.

Required by cyber4all/Clark-client#711

davidnso:
Changes in this PR include... 

- Adding the optional `hasrevision` flag to learning objects coming down in searches by privileged users. 
- implement an aggregation to check whether or not an object has a revision when navigating to the details page. 
- remove `reportError` call from the map that loads learning object metrics.

closes #277


src/drivers/MongoDriver.ts
```diff
@@ -1291,7 +1314,7 @@ export class MongoDriver implements DataStore {
   }
 
   /**
-   * Fetches released object
+   * Fetches released object through aggregation pipeline.
```
DonSeannelly: Fetches a released Learning Object, noting whether the working copy of that object has revisions made to it by doing a comparison of the status property on both documents.

PR #308, opened 2019-04-05
Reconfigure npm start scripts to use nodemon and tsc

gshaw1997:
This PR updates start scripts to use tsc-node and nodemon instead of gulp to start the development server. This fixes the issue gulp and nodemon caused which forced unnecessary nodemon restarts in file change.


gshaw1997:
> Config updates look good, but I'm not following the changelog and `SubmissionDatastore` updates in here?

There imports for these were wrong because of typos. This caused tsc build to fail. There not throwing errors in production because they are just interfaces

PR #309, opened 2019-04-05
improve in memory mongo database structure

### LINKED ISSUE #266: Improve in-memory-mongodb seed data
With tests becoming more of a priority in our current workflow, as we transition from using mocks to utilizing the in-memory-mongodb server for unit and endpoint tests, it would be beneficial for seed data stored in-memory to match data stored in our dev and production database.

davidnso:
This PR reorganized the data in our seed database and extracts the shared Id's between objects to allow us to reference those objects in the database a bit easier when writing tests.

Changes to actual seed data: 

- Learning objects with different statuses; Waiting, Proofing, and Review. 

- A released Parent object and child now have a working copy

- The released parent object has a released and waiting child object. 

- New privileged users; Admin, Editor, reviewer@nccp, curator@nccp

closes #266


davidnso:
> Seed data looks good.
> 
> The expect statements within the endpoint tests should be updated to validate against the response contract.

I didn't change anything with the endpoint tests here since it was out of scope for this PR. The other PR on endpoint tests would handle that though.

PR #311, opened 2019-04-09
Update released counter to count docs in released collection

gshaw1997:
This PR fixes the discrepancy between the released counter on the home page versus the object count on the browse page. The released counter is now populated with the number of documents within the released-objects collection instead of the amount of objects with a status of `released` within the objects collection.


gshaw1997:
> Why is that number different to begin with? Shouldn't they be the same?

@nrwinner It appears that the status of a working copy learning object changed from released to some unreleased status, so at that point the count was off.

src/LearningObjectStats/LearningObjectStatStore.ts
```diff
@@ -26,14 +26,17 @@ export class LearningObjectStatStore implements LearningObjectStatDatastore {
    * @memberof LearningObjectStatStore
    */
   async fetchStats(params: { query: any }): Promise<LearningObjectStats> {
+    const releasedCount$ = this.db
+      .collection(COLLECTIONS.RELEASED_LEARNING_OBJECTS)
+      .find({})
+      .count();
```
DonSeannelly: Could we use the `count()` method directly on the collection? I believe it is more performant as it doesn't require a cursor to be created.

This SO answer covers it: https://stackoverflow.com/a/45000351/6378660

gshaw1997: Good catch

PR #312, opened 2019-04-09
Create Learning Object Submission Collection

nvisal1:
The purpose of this PR is to record all learning object submissions in a Mongo collection. 

Docs: https://documenter.getpostman.com/view/4179920/S1EMVf3i

This PR changes the structure of existing submission routes to follow `/users/:userId`. Because of this, gateway and client updates are required.


src/LearningObjectSubmission/AuthorizationManager.ts
```diff
@@ -0,0 +1,26 @@
...
+    learningObjectId: string,
+  }): Promise<void> {
+    if (!params.emailVerified) {
+      throw new ResourceError(
+        'Invalid Access',
+        ResourceErrorReason.INVALID_ACCESS,
+      );
+    }
+
+    if (!(await params.dataStore.checkLearningObjectExistence({
```
DonSeannelly: I think the code would read a lot better if we made this function call outside of the if statement, and used a variable. There's just a lot happening inside of the if statement here.

src/LearningObjectSubmission/SubmissionDatastore.ts
```diff
@@ -25,27 +26,65 @@ export class SubmissionDatastore {
...
+    await this.db.collection(COLLECTIONS.SUBMISSIONS)
+      .findOneAndUpdate(
+        { learningObjectId },
         {
-          $set: {
-            published: true,
-            status: 'waiting',
-            collection,
-          },
+          $set: { cancelDate: Date.now().toString() },
```
DonSeannelly: Are there any noticeable differences between storing this as a string, vs Mongo's actual timestamp?

src/LearningObjectSubmission/SubmissionRouteDriver.ts
```diff
@@ -24,41 +26,74 @@ export function initialize({
...
+  async function fetchSubmission(req: Request, res: Response) {
+    try {
+      const learningObjectId = req.params.learningObjectId;
+      const userId = req.params.userId;
+      const collection = req.query.collection;
+      const hasSubmission = req.query.hasSubmission;
+      const emailVerified = req.user.emailVerified;
+
+      if (!collection || !hasSubmission) {
+        res.status(501).json({message: 'Not Implemented'});
```
DonSeannelly: Something like this might make debugging easier on the client-side:

```suggestion
        res.status(501).json({message: 'This route currently requires both the collection and hasSubmission query parameters.'});
```

src/LearningObjectSubmission/SubmissionRouteDriver.ts
```diff
@@ -24,41 +26,74 @@ export function initialize({
...
+      const { code, message } = mapErrorToResponseData(e);
+      res.status(code).json({message});
     }
   }
 
-  router.post('/learning-objects/:learningObjectId/submission', submit);
-  router.delete('/learning-objects/:learningObjectId/submission', cancel);
+  router.get('/users/:userId/learning-objects/:learningObjectId/submission', fetchSubmission);
+  router.post('/users/:userId/learning-objects/:learningObjectId/submission', submit);
+  router.delete('/users/:userId/learning-objects/:learningObjectId/submission', cancel);
```
DonSeannelly: Let's make submission plural (as discussed in person)

src/LearningObjectSubmission/SubmissionDatastore.ts
nvisal1: I used a string here to keep timestamps consistent throughout the system. Other timestamps use `Date.now().toString()`.

PR #314, opened 2019-04-15
Allow cancellation of unrecorded submissions

nvisal1:
This PR removes a check in the `cancelSubmission` function to allow for unrecorded submissions to be canceled.


gshaw1997:
@nvisal1 We should make a note to undo the removal of this check, once data has been appropriately back filled.

PR #316, opened 2019-04-22
Allow addition of multiple file meta

gshaw1997:
This PR updates the POST route for `/materials/files` to allow client to add metadata for multiple files with a single request.

### DIFF
--- a/src/LearningObjects/LearningObjectInteractor.ts
+++ b/src/LearningObjects/LearningObjectInteractor.ts
@@ -52,7 +52,7 @@ const LearningObjectState = {
  * @param {string} authorUsername [Learning Object's author's username]
  * @param {string} learningObjectId [Id of the Learning Object to add the file metadata to]
  * @param {FileMeta} fileMeta [Object containing metadata about the file]
- * @returns {Promise<string>} [Id of the file metadata]
+ * @returns {Promise<string>} [Id of the added Learning Object file]
  */
 export async function addLearningObjectFile({
   dataStore,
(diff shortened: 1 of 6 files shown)


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -476,15 +516,17 @@ export async function deleteLearningObject(params: {
...
-            `Problem deleting changelogs for ${
-              params.learningObjectName
-            }: ${e}`,
-          ),
-        );
-      });
+      params.dataStore
+        .deleteChangelog({ learningObjectId: object.id })
+        .catch(e => {
+          reportError(
```
DonSeannelly: It looks like just formatting was touched here, but it looks like this doesn't return any error state from the interactor scope. @nvisal1 can you check on this? It may cause the service to send a success response even if we error out.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -95,6 +95,46 @@ export async function addLearningObjectFile({
...
+}: {
+  dataStore: DataStore;
+  requester: UserToken;
+  authorUsername: string;
+  learningObjectId: string;
+  fileMeta: FileMeta[];
+}): Promise<string[]> {
+  try {
+    const promises$ = fileMeta.map(file => {
+      return addLearningObjectFile({
```
DonSeannelly: Does this function enforce the authorization rules listed in the JSDoc? I'm not seeing any logic inside of here.

gshaw1997: The `addLearningObjectFile` that's is being called for each file enforces authorization rules

DonSeannelly: Is it worth having the JSDoc for this function that specifies enforcement of those rules if the function itself doesn't actually do it? I think I could lean either way, I just bring it up because I was confused looking at this function in isolation.

gshaw1997: I see. Makes sense to remove it if it is confusing. The docs should represent how the function operates in isolation.

src/LearningObjects/LearningObjectInteractor.ts
gshaw1997: It seems as though we are allowing changelog deletions to fail and reporting those failures. The actual deletion of the learning object does either complete or reject an error.

PR #317, opened 2019-04-22
Update change log structure in database

nvisal1:
This PR changes the structure of change logs in the database in order to minimize the number of client requests needed to render the change log UI in the dashboard. 

This PR adds the `name`, `role`, and `profileImage` properties to change log documents. 
Existing change log documents will need to be updated when merging this PR.


src/Changelogs/AuthManager.ts
```diff
@@ -50,10 +28,31 @@ async function userIsOwner(params: {
...
-        accessGroups.includes(ROLE.EDITOR))
+export async function determineRole(params: {
+    dataStore: DataStore,
+    user: UserToken,
+    learningObjectId: string,
+    accessGroups: string[],
+}): Promise<string> {
+
+    const userId = await params.dataStore.findUser(params.user.username);
+    const learningObject = await params.dataStore.checkLearningObjectExistence({
```
gshaw1997: Does this datastore call return the full Learning Object?

src/Changelogs/AuthManager.ts
```diff
@@ -12,36 +14,12 @@ export async function hasChangelogAccess(params: {
...
-  }) {
-    const userId = await params.dataStore.findUser(params.user.username);
-    if (await params.dataStore.checkLearningObjectExistence({
-            userId,
-            learningObjectId: params.learningObjectId,
-        })) {
-            return true;
-        }
-    return false;
+    return await determineRole({
```
DonSeannelly: Why does this function proxy to another? What's the design reason for not just calling determineRole directly?

src/Changelogs/AuthManager.ts
DonSeannelly: By the name of the function, I would assume a boolean return type. It also seems to be used here to more so identify the author's ownership, not anything specific about the Learning Object. Maybe a variable name similar to `isOwnedByAuthor` would be a better fit?

src/Changelogs/ChangelogInteractor.ts
```diff
@@ -106,25 +114,37 @@ async function authorizeRequest(params: {
...
-  }))) {
-    throw new ResourceError(
-      'Invalid Access',
-      ResourceErrorReason.INVALID_ACCESS,
-    );
-  }
+  });
 
-  if (!(await params.dataStore.checkLearningObjectExistence({
+  const learningObject = await params.dataStore.checkLearningObjectExistence({
```
DonSeannelly: Similar to the discussion above

src/Changelogs/ChangelogInteractor.ts
```diff
@@ -106,25 +114,37 @@ async function authorizeRequest(params: {
...
   }
+
+  return role;
+}
+
+function generateProfileImageUrl(params: {
+  email: string,
+}): string {
+  const defaultIcon = 'identicon';
+  // r=pg checks the rating of the Gravatar image
```
DonSeannelly: What does this mean? Could we perhaps move it to a more detailed JSDoc comment for this function?

PR #320, opened 2019-05-24
Add an aggregation step to order files by date

nvisal1:
This PR addresses story #226
https://app.clubhouse.io/clarkcan/story/226/the-files-in-the-file-viewer-in-the-details-page-should-be-sorted-by-last-updated-newest-first

The purpose of this PR is to sort files by date so that they are displayed newest first on the details page. 

Screenshot: 
Screen Shot 2019-05-23 at 9 58 54 PM

I am not a huge fan of the syntax used in this aggregation. If anyone has a more succinct way to write this, please let me know. The biggest problem that I faced when writing this aggregation was caused by MongoDB's inability to group by nested fields. Because of this, I could not directly push the sorted array to `materials.files`. Instead, I added a new field to the object and performed an assignment in memory.


src/drivers/MongoDriver.ts
```diff
@@ -1385,19 +1391,58 @@ export class MongoDriver implements DataStore {
...
-      .findOne({ _id: params.id });
-    if (object) {
-      const author = await this.fetchUser(object.authorID);
-      return this.generateLearningObject(author, object, params.full);
+      .aggregate([
+        {
+          // match learning object by params.id
+          $match: { _id: params.id },
+        },
+        { $unwind: '$materials.files' },
```
gshaw1997: To avoid having to make a second query in the case were `materials.files` is an empty array or null, use the `preserveNullAndEmptyArrays` option for the `$unwind` operator.
https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/

```suggestion
        { $unwind: {path: '$materials.files', preserveNullAndEmptyArrays: true} },
```

src/drivers/MongoDriver.ts
```diff
@@ -1416,6 +1461,27 @@ export class MongoDriver implements DataStore {
           // match learning object by params.id
           $match: { _id: params.id },
         },
+        { $unwind: '$materials.files' },
```
gshaw1997: To avoid having to make a second query in the case were `materials.files` is an empty array or null, use the `preserveNullAndEmptyArrays` option for the `$unwind` operator.
https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/

```suggestion
        { $unwind: {path: '$materials.files', preserveNullAndEmptyArrays: true} },
```

src/drivers/MongoDriver.ts
```diff
@@ -1424,13 +1422,6 @@ export class MongoDriver implements DataStore {
       const author = await this.fetchUser(object[0].authorID);
```
gshaw1997: We should check to make sure an object was found before proceeding.

src/drivers/MongoDriver.ts
```diff
@@ -1511,41 +1500,6 @@ export class MongoDriver implements DataStore {
...
```
gshaw1997: We should check to make sure an object was found before proceeding.

PR #322, opened 2019-05-24
Create Singleton Mongo Connection Host

DonSeannelly:
This PR takes the pattern of using a Singleton to hold an instance of `MongoClient` that independent modules within the service can share. This was introduced in other services, however there is a slight tweak here.

In order to prevent refactoring of the current `MongoDriver` and keep tests green, I still have the `MongoDriver` run its build method as normal in the `app.ts`. In that build method (this is to keep tests passing) I added a small conditional that if the `MongoConnector` hasn't opened its connection yet, the `MongoDriver` will ask it to do so - then proceed as normal.

I have left FIXMEs in the code on the areas that we should make sure we go back and update after we've continued down the road of refactoring.

### DIFF
--- a/src/shared/Mongo/MongoConnector.ts
+++ b/src/shared/Mongo/MongoConnector.ts
@@ -60,7 +60,7 @@ export class MongoConnector {
    * and scripts using the database, so only do this if it's very
    * important or if you are sure that *everything* is finished.
    */
-  disconnect(): void {
-    MongoConnector.mongoClient.close();
+  static disconnect(): Promise<void> {
+    return MongoConnector.mongoClient.close();
   }
 }
--- a/src/shared/Mongo/MongoConnector.ts
+++ b/src/shared/Mongo/MongoConnector.ts
@@ -1,0 +1,66 @@
+import { MongoClient, Db } from 'mongodb';
+import { reportError } from '../../drivers/SentryConnector';
+import { ServiceError, ServiceErrorReason } from '../errors';
+
+export class MongoConnector {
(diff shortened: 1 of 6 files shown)


src/shared/Mongo/MongoConnector.ts
```diff
@@ -0,0 +1,66 @@
...
+    return this.mongoClient;
+  }
+
+  /**
+   * Connect to the database. Must be called before any other functions.
+   * @param {string} dbIP the host and port on which mongodb is running
+   */
+  private async connect(dbURI: string, retryAttempt?: number): Promise<void> {
+    try {
+      MongoConnector.mongoClient = await MongoClient.connect(dbURI);
```
gshaw1997: We should use MongoClient's baked in retry mechanism instead of the home baked one.
https://mongodb.github.io/node-mongodb-native/api-generated/mongoclient.html

```suggestion
      MongoConnector.mongoClient = await new MongoClient(dbURI, {reconnectTries: 1).connect();
```

src/shared/Mongo/MongoConnector.ts
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Retrieves the client with an open connection to the database.
+   */
+  static client(): MongoClient {
+    return this.mongoClient;
+  }
+
+  /**
+   * Connect to the database. Must be called before any other functions.
+   * @param {string} dbIP the host and port on which mongodb is running
```
gshaw1997:
```suggestion
   * @param {string} dbURI the host and port on which mongodb is running
```

PR #329, opened 2019-05-28
Test Updates that Release Learning Objects

DonSeannelly:
This PR adds test cases to assert that when a Learning Object update request is made by an editor or an admin with `status: released` included, the service responds with a 200 OK.

### DIFF
--- a/src/LearningObjects/LearningObjectRouteHandler.spec.ts
+++ b/src/LearningObjects/LearningObjectRouteHandler.spec.ts
@@ -103,10 +103,7 @@ describe('LearningObjectRouteHandler', () => {
             });
             describe('and the requester is an editor', () => {
                 it('should update the requested Learning Object and return a status of 200', done => {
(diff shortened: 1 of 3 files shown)


src/LearningObjects/LearningObjectRouteHandler.spec.ts
```diff
@@ -85,6 +87,39 @@ describe('LearningObjectRouteHandler', () => {
...
+                        .then(res => {
+                            done();
+                        },
+                        );
+                });
+            });
+            describe('and the requester is an editor', () => {
+                it('should update the requested Learning Object and return a status of 200', done => {
+                    const editorToken = generateToken({
+                        ...MOCK_OBJECTS.USERTOKEN,
```
davidnso: There is a `USERTOKEN_EDITOR` object in the mocks file that you can use for this to make things easier.

PR #330, opened 2019-05-28
Create Publishing Module

DonSeannelly:
This PR moves the logic of publishing a LearningObject (aka "releasing") to its own module, rather than including it inside of the existing `LearningObjectInteractor` and `MongoDriver`.

### Approach
The Publishing module exports an anonymous function that instantiates a new `ReleaseRequestDuplicator` (I'm not sold on the name here, so any suggestions are welcome) and replaces the `dataStore` provided to the interactor function, `releaseLearningObject`, with an instance of this class instead.

Then internally, the `ReleaseRequestDuplicator` makes a request to index the Learning Object into ElasticSearch before proceeding as normal to call `addToReleased` on the `MongoDriver` that is hidden behind service-wide `DataStore` interface.

### The Name Choice
I chose *Publishing* module based on a conversation we had about Domain Driven Design models for our system where we discussed a Publishing Department. I thought that this Publishing Department would make a good standalone service one day, and so a good place to start would be with a module inside of the Learning Object Service right now.

### Other Changes
Additionally, as a side-effect of working through the release authorization again, I noticed that we were using `AuthorizationManager#isPrivilegedUser` instead of `AuthorizationManager#isAdminOrEditor` when checking if a user was able to release a Learning Object. This meant that in the case that an author of a Learning Object for a collection was also a reviewer in that same collection, they could release their own Learning Object.

### DIFF
--- a/src/LearningObjects/Publishing/ReleaseRequestDuplicator.ts
+++ b/src/LearningObjects/Publishing/ReleaseRequestDuplicator.ts
@@ -1,0 +1,33 @@
+import { PublishingDataStore } from './interactor';
+import { LearningObject } from '../../shared/entity';
+import { reportError } from '../../shared/SentryConnector';
+import { ElasticSearchPublishingGateway } from './ElasticSearchPublishingGateway';
+
+/**
+ * This class is an abstraction point that allows us to use Mongo and
+ * ElasticSearch side-by-side. It is written with the intent of ElasticSearch
+ * running in "dark mode", where we index data into it but do not read data
+ * out of it.
+ */
+export class ReleaseRequestDuplicator implements PublishingDataStore {
+  elasticSearchStore: PublishingDataStore;
+  constructor(private mongoStore: PublishingDataStore) {
+    this.elasticSearchStore = new ElasticSearchPublishingGateway();
+  }
+
+  /**
+   * Tries to index the Learning Object into ElasticSearch, then proceeds
+   * to make the call to Mongo as normal. If the request to ElasticSearch
+   * fails, we report the error and proceed as if nothing happened.
+   *
+   * @param releasableObject the Learning Object that should be released
+   */
+  async addToReleased(releasableObject: LearningObject): Promise<void> {
+    try {
+      await this.elasticSearchStore.addToReleased(releasableObject);
+    } catch (e) {
+      reportError(e);
+    }
+    return this.mongoStore.addToReleased(releasableObject);
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "learning-object-service",
-  "version": "2.18.4-1",
+  "version": "2.18.4-2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "learning-object-service",
-  "version": "2.18.4-1",
+  "version": "2.18.4-2",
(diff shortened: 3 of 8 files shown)


src/LearningObjects/Publishing/ReleaseRequestDuplicator.ts
```diff
@@ -0,0 +1,33 @@
...
+   *
+   * @param releasableObject the Learning Object that should be released
+   */
+  async addToReleased(releasableObject: LearningObject): Promise<void> {
+    try {
+      await this.elasticSearchStore.addToReleased(releasableObject);
+    } catch (e) {
+      reportError(e);
+    }
+    return this.mongoStore.addToReleased(releasableObject);
```
davidnso: if the mongostore is of type `PublishingDatastore` how is it getting access to the mongo-driver directly?

src/LearningObjects/Publishing/ReleaseRequestDuplicator.ts
```diff
@@ -0,0 +1,33 @@
...
+import { reportError } from '../../shared/SentryConnector';
+import { ElasticSearchPublishingGateway } from './ElasticSearchPublishingGateway';
+
+/**
+ * This class is an abstraction point that allows us to use Mongo and
+ * ElasticSearch side-by-side. It is written with the intent of ElasticSearch
+ * running in "dark mode", where we index data into it but do not read data
+ * out of it.
+ */
+export class ReleaseRequestDuplicator implements PublishingDataStore {
```
davidnso: I'd suggest a more general name for this like `ElasticMongoBridge` or something along those lines. If we want to keep it specific then the current name is fine.

src/LearningObjects/Publishing/ReleaseRequestDuplicator.ts
DonSeannelly: `PublishingDatastore` is a subtype of `DataStore`, which `MongoDriver` implements. Type-wise, it doesn't think it has access, but the actual instantiated class that is being passed in is `MongoDriver`.

src/LearningObjects/Publishing/ReleaseRequestDuplicator.ts
DonSeannelly: Can you make a case for generalizing it?  I could be convinced either way, I think. My thoughts now are that we will probably remove this component from the service before the need to expand it for other use cases comes up.

davidnso: I'd say the general name gives a good description on the two datastore's the class in communicating with. When I see `ElasticMongoBridge` implements `PublishingDataStore` , at first glance I have a better idea of what's going on in the class as a whole. Especially if we're going to expand it for other use cases

DonSeannelly: I think I see your point. I think my perspective is differing because I don't believe that we should be expanding this class. It is just a temporary abstraction that was put in place with the intent to remove it, rather than a permanent piece of the system. With that, I think a generic class name reduces the ability to comprehensively understand its purpose.

DonSeannelly: @davidnso I talked with Gus a little on this. Would you prefer something like `ElasticMongoReleaseRequestDuplicator`? It then includes the data store types, but also fits my bill of not becoming generic.

davidnso: Yea that works

PR #333, opened 2019-06-03
Extract cleanDocument to shared

DonSeannelly:
This PR moves the cleanDocument function to a new shared module `elasticsearch/HelperFunctions`, as well as renames it to cleanLearningObject.

This is done in preparation for indexing submissions into Elasticsearch in addition to releases.


src/LearningObjects/Publishing/ElasticSearchPublishingGateway.ts
```diff
@@ -16,29 +17,7 @@ export class ElasticSearchPublishingGateway implements PublishingDataStore {
       headers: {
         'Content-Type': 'application/json',
```
nvisal1: Add host header to avoid aws invalid host header errors. The environment variable can be found in outcome suggestion service.

PR #335, opened 2019-06-04
Remove false positive tests

DonSeannelly:
This PR removes tests that are wrongfully passing due to bad logic in the tests or poorly configured test environment conditions.

After talking with @nvisal1, we also realized that these tests are make assertions about the helper functions they use, not the logic they implement, and therefore should be rewritten. This PR only removes the bad tests, and future work with our test environment will go back to implement new, better tests.

### DIFF
--- a/src/LearningObjectSubmission/SubmissionInteractor.spec.ts
+++ b/src/LearningObjectSubmission/SubmissionInteractor.spec.ts
@@ -1,100 +1,24 @@
-import { submitForReview, cancelSubmission } from './SubmissionInteractor';
+import { cancelSubmission } from './SubmissionInteractor';
 import {
   MOCK_OBJECTS,
   SUBMITTABLE_LEARNING_OBJECT,
 } from '../tests/mocks';
 import { DataStore } from '../shared/interfaces/DataStore';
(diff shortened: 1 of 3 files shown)


src/LearningObjectSubmission/SubmissionInteractor.spec.ts
```diff
@@ -1,100 +1,24 @@
...
-          userId: MOCK_OBJECTS.USER_ID,
-          emailVerified: true,
-          learningObjectId: SUBMITTABLE_LEARNING_OBJECT.id,
-        }),
-      ).resolves.toBe(undefined);
-      done();
-    } catch (error) {
-      console.log(error);
-    }
+    await expect(
```
jbarlow96: What was the point of removing the try/catch here? Just for clarification.

DonSeannelly: We want the test to pass without any errors. With the try-catch, if the test would throw an error (which means logically that something went wrong and it should fail) we would catch it and ignore it. This would allow a "failing" test to still pass, with the only difference being an error lost to the console.

PR #337, opened 2019-06-04
Refactor Learning Object Search to Module

gshaw1997:
This PR refactors all biz logic related to performing a search on learning objects into a module.

 Contributes to https://app.clubhouse.io/clarkcan/story/265/perform-search-via-elasticsearch-indices.


src/LearningObjectSearch/Interactor.ts
```diff
@@ -0,0 +1,238 @@
...
+} from '../shared/functions';
+import {
+  requesterIsPrivileged,
+  requesterIsAdminOrEditor,
+} from './AuthorizationManager';
+import { LearningObjectSearchGateway } from './interfaces';
+import { getAccessGroupCollections } from '../shared/AuthorizationManager';
+import { ResourceErrorReason, ResourceError } from '../shared/errors';
+
+namespace Gateways {
```
DonSeannelly: Why do we use a namespace here?

src/LearningObjectSearch/index.ts
```diff
@@ -0,0 +1,14 @@
...
+  ExpressServiceModule,
+  expressServiceModule,
+} from 'node-service-module';
+import { LearningObjectSearchGateway } from './interfaces';
+import { ExpressHttpAdapter } from './adapters';
+
+@expressServiceModule({
+  expressRouter: ExpressHttpAdapter.buildRouter(),
+  providers: [
+    { provide: LearningObjectSearchGateway, useClass: null },
```
DonSeannelly: Can you explain what this is doing? My understanding would be that it is providing a LearningObjectSearchGateway to this module, which is the LearningObjectSearch module. I'm a little confused on that, and also why `useClass` is set to `null`

src/LearningObjectSearch/interfaces/LearningObjectDatastore.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  ReleasedLearningObjectSearchQuery,
+  LearningObjectSearchQuery,
+  LearningObjectSearchResult,
+} from '../typings';
+
+export abstract class LearningObjectSearchGateway {
```
DonSeannelly: Why is the `LearningObjectSearchGateway` abstract class located in the `LearningObjectDatastore` interface file?

src/shared/functions/index.ts
```diff
@@ -40,3 +40,44 @@ export function sanitizeObject<T>(
   }
   return params.object;
 }
+
+/**
+ * Returns new array with element(s) from value param or undefined if value was not defined
```
DonSeannelly: It looks like line 53 returns null explicitly, instead of undefined

```suggestion
 * Returns new array with element(s) from value param or null if value was not defined
```

src/shared/functions/index.ts
```diff
@@ -40,3 +40,44 @@ export function sanitizeObject<T>(
...
+  if (value == null || value === '') {
+    return null;
+  }
+  if (value && Array.isArray(value)) {
+    return [...value].filter(v => !isEmptyValue(v));
+  }
+  return [value].filter(v => !isEmptyValue(v));
+}
+
+function isEmptyValue(value: any): boolean {
```
DonSeannelly: This function is missing JSDoc

src/shared/functions/index.ts
```diff
@@ -40,3 +40,44 @@ export function sanitizeObject<T>(
...
+/**
+ *
+ * Converts non-number value to number if defined else returns null
+ * @param {*} value
+ * @returns {number}
+ */
+export function toNumber(value: any): number {
+  const num = parseInt(`${value}`, 10);
+  if (!isNaN(num)) {
+    return +value;
```
DonSeannelly: Why return `+value` here instead of num?

src/LearningObjectSearch/Interactor.ts
gshaw1997: Its used as a method organize the available Gateways, Drivers, etc as an internal module. As/If we add more, we’re going to want to have some kind of organization scheme so that we can keep track of our types and not worry about name collisions with other objects. Instead of putting lots of different names into the global namespace, I chose to wrap up the objects into a namespace.

src/LearningObjectSearch/index.ts
gshaw1997: The `LearningObjectSearchGateway` is an interface that any of our SearchGateways/Datastore will implement. I now realize the Gateway naming may be confusing, but my rationale for choosing Gateway instead of Datastore was that Datastore seemed to imply that it would/could be used to store data and have CUD capabilities when in actuality it is only proxying Read requests to an external service.


Naming aside, yes this would be providing a `LearningObjectGateway` to this module to be used at the interactor level. I should have noted this, but it is `null` now because this PR just handled refactoring of biz-logic into a module, so no driver implementation is available to provide. Once the Elastic Search Driver/Gateway is implemented, that will be the class that is provided.

src/LearningObjectSearch/interfaces/LearningObjectDatastore.ts
gshaw1997: I believe this is noted in some of the earlier PR's in other service that utilize this pattern, but it is an pure abstract class that serves as an interface. This is so that the class's name can be used as a value for the DI provider tokens which isn't possible with interfaces because they have no run time artifact. 

Using pure abstract classes as interfaces, so that the name could be used as the DI token made it easier to define, provide, and request dependencies because there is no need to define and manage extra const values to serve as the injection keys.

Like much of this module pattern, this decision was also heavily influenced by Angular's DI which discusses using abstract classes to circumvent Typescript's interface limitations [here](https://angular.io/guide/dependency-injection-in-action#provider-token-alternatives-class-interface-and-injectiontoken).

src/LearningObjectSearch/index.ts
DonSeannelly: > Naming aside, yes this would be providing a LearningObjectGateway ...

Is the provide statement supposed to say `LearningObjectGateway`, or did you mean to say `LearningObjectSearchGateway` in the above statement?

Based on the other writing, I'm assuming the code is written correctly and the type-o is in that comment. If that is the case, this becomes kind of interesting. I see your point on Datastore vs. Gateway here. However, from our discussions, this naming kind of muddies the conventions since now this Gateway is an interface for direct data access, rather than business logic functions that in turn own their own data store.

I think I'd like to advocate for reserving the term `Gateway` for a request out to another unit of business logic execution that can in turn work with data. Just to keep it a single-purpose term. Thoughts? I'm sort of going back and forth on this.

> I should have noted this, but it is null now ...

Can we add a fixme comment here?

src/LearningObjectSearch/interfaces/LearningObjectDatastore.ts
DonSeannelly: Thanks for the link! I see how the abstract class is useful for DI here. I am still curious as to the convention here in that the file name is not consistent with the member it exports. What is the reason for that?

src/LearningObjectSearch/Interactor.ts
DonSeannelly: I have no real experience with namespaces (default TSLint rules for most projects have discouraged them). Does this mean this namespace is global to the application, or just this module?

src/LearningObjectSearch/AuthorizationManager.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+    it('should return true when accessGroups contain all privileges', () => {
+      requester.accessGroups = [
+        'admin',
+        'editor',
+        'curator@collection',
+        'reviewer@collection',
+      ];
+      expect(auth.requesterIsPrivileged(requester)).toBe(true);
+    });
+    it('should return fale when accessGroups contain no privileges', () => {
```
DonSeannelly:
```suggestion
    it('should return false when accessGroups contain no privileges', () => {
```

src/LearningObjectSearch/index.ts
gshaw1997: Yes, comment was a typo.

I was going back and fourth as well which is why the filename and type that is exported is not consistent; ended up forgetting to delete. For the sake of reducing confusion I am okay with reserving the `Gateway` naming convention for a request out to another unit of business logic execution. 

Any suggestions on interface name here though? Does `Datastore` accurately describe the interface here?

src/LearningObjectSearch/interfaces/LearningObjectDatastore.ts
gshaw1997: The file name should have been renamed to `LearningObjectSearchGateway` to match the class name, it was just missed during my internal debate of naming conventions.

src/LearningObjectSearch/Interactor.ts
gshaw1997: It is only global to the module, unless exported.

src/LearningObjectSearch/index.ts
DonSeannelly: @gshaw1997 I think that it can make sense for this. If you think of it purely as "a place for data storage" it doesn't necessarily mean that new data can be written/updated (even though most of our existing `DataStore`s have full CRUD). Some storage is read-only, and that's what we have here.

I think that in terms of architectural level, `DataStore` makes sense because it's indicative of this being a plug-in for some time of direct data operation, rather than some business logic execution.

(As I'm writing this I'm becoming more convinced that calling it a DataStore is a good idea haha)

PR #338, opened 2019-06-05
Add Lambda Function Invocation For Sending Release Email

nvisal1:
This PR adds a lambda function invocation (HTTP request) that triggers the lambda function responsible for dispatching learning object release notification emails. This request is fired when a learning object is released.

The lambda trigger is located in the `LambdaGateway` class. This gateway is local to the `Publishing` module. A "fake" lambda trigger is located in the `LambdaGatewayStub` class. Both the `LambdaGateway` and `LambdaGatewayStub`  implement the `ReleaseEmailGateway` interface. During the execution of the `Publishing` module index, the lambda gateway instance is set by the `LambdaGatewayFactory` class. The class determines whether to create an instance of `LambdaGatewayStub` or `LambdaGateway` by looking at the `NODE_ENV` environment variable. 

`testing || development --> LambdaGatewayStub`
`production --> LambdaGateway`
`undefined || empty string --> internal service error`

This prevents developers from sending accidental emails during development and allows us to use the stub when running tests.

A new env is introduced in this PR. `RELEASE_EMAIL_INVOCATION` is the lambda endpoint. 

In order to access the function, the request must contain a bearer token that has the signature of a ServiceToken sent from Learning Object Service. 

A new function was also added to the global `errors.ts` file. The function `mapResponseDataToError` allows us to interpret status codes returned from external services.


Dockerfile
```diff
@@ -40,6 +40,7 @@ ENV PATH /opt/node_modules/.bin:$PATH
 
 # Swtich working dir to opt to use node_modules for testing
 WORKDIR /opt
+ARG NODE_ENV=development
```
nvisal1: This line sets the `NODE_ENV` to development before running tests. The `NODE_ENV` is set to production after tests complete. Prior to this update, we were not setting the `NODE_ENV` before running tests.

DonSeannelly: Does this then fix the issue that we discussed about running tests in Circle CI?

src/shared/errors.ts
```diff
@@ -77,3 +77,42 @@ export function mapErrorToResponseData(error: Error): { code: number, message: s
...
+        ResourceErrorReason.FORBIDDEN,
+      );
+    case 404:
+      return  new ResourceError(
+        message,
+        ResourceErrorReason.NOT_FOUND,
+      );
+    default:
+      return new ServiceError(ServiceErrorReason.INTERNAL);
+  }
```
DonSeannelly: The way that this is being used would mean that our API would leak the error messages from whatever upstream source provided an error.  This should be a detail that we handle at a lower level, and throw an appropriate error message depending on what the conditional logic dictates. 

That way we don’t have to worry about AWS lambda function error messages making their way to our client.

src/LearningObjects/Publishing/lambda-gateway.ts
```diff
@@ -0,0 +1,56 @@
...
+                'Content-Type': 'application/x-www-form-urlencoded',
+                'Content-Length': postData.length,
+                'Authorization': `Bearer ${generateServiceToken()}`,
+            },
+        };
+
+        const req = https.request(options, (res) => {
+            if (res.statusCode !== 200) {
+                const error = mapResponseDataToError(res.statusCode, res.statusMessage);
+                throw error;
```
DonSeannelly: I don’t think we should error out the operation because the email failed to send. I think it would be better instead of throwing this error and preventing the route from responding correctly, it might be better to just report this error to sentry so that we can fix it retrospectively

src/LearningObjects/Publishing/lambda-gateway-factory.ts
```diff
@@ -0,0 +1,38 @@
...
+import { LambdaGateway } from './lambda-gateway';
+
+export class LambdaGatewayFactory {
+
+    static instance: ReleaseEmailGateway;
+    private constructor() {}
+
+    /**
+     * Returns an instance of ReleaseEmailGateway
+     * Follows a singleton pattern to ensiure only instance exists
```
DonSeannelly: Why does this need to be a singleton, when the actual implementation of the release email gateway has no state? Any part of the system could just instantiate a new member with no side effects.  Maybe I am not fully understanding this piece

nvisal1: I implemented this as a singleton because of our meeting yesterday morning. We decided that each gateway would be stored as a singleton instance within the gateway registry.

src/shared/errors.ts
nvisal1: Good catch

Dockerfile
nvisal1: Yes it does

src/LearningObjects/Publishing/lambda-gateway-factory.ts
DonSeannelly: Ah, I understand now.  Based on work done today, and some conversations with Gus, I think that we may have oversimplified our statements yesterday. My opinion on this is that since this is something that is specific to the release process, meaning that we don’t really want to open this up for anybody else to reach out to, it would probably be best to just keep this as a simple class inside of the internal workings of this module. 

I don’t think we’re at a point where this is a science yet, so this is definitely just an opinion of mine. I’m interested in what others think on this? @gshaw1997 @davidnso @nvisal1

davidnso: I agree with Sean, making it a simple class within the module would make more sense since sending out an email is a part of the release process.

src/LearningObjects/Publishing/index.ts
```diff
@@ -12,8 +12,8 @@ const setupElasticToggle = ({ userToken, dataStore, releasableObject }: {
   releasableObject: LearningObject;
 }) => {
   const toggle = new ElasticMongoReleaseRequestDuplicator(dataStore);
-  const releaseEmailGateway = LambdaGatewayFactory.getLambdaGatewayInstance();
-  releaseLearningObject({ userToken, dataStore: toggle, releasableObject, releaseEmailGateway });
+  const releaseEmailGateway = new LambdaGatewayFactory();
+  releaseLearningObject({ userToken, dataStore: toggle, releasableObject, releaseEmailGateway: releaseEmailGateway.instance });
```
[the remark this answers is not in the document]
DonSeannelly: Actually, I think I answered my own question. The `new` keyword is considered an anti-pattern for factory classes, because it must return the exact type. Look into creating a function that returns a `ReleaseEmailGateway`. Perhaps `LambdaGatewayFactory.buildGateway()`? You can read more about the fun world of factories here https://sourcemaking.com/design_patterns/factory_method

src/LearningObjects/Publishing/lambda-gateway-factory.ts
```diff
@@ -5,34 +5,31 @@ import { LambdaGateway } from './lambda-gateway';
...
+                this._instance = new LambdaGatewayStub();
+                break;
+            case 'development':
+                this._instance = new LambdaGatewayStub();
+                break;
+            case 'production':
+                this._instance = new LambdaGateway();
+                break;
+            default:
+                throw new ServiceError(ServiceErrorReason.INTERNAL);
```
DonSeannelly: This logic could be moved into the build function I describe above

src/LearningObjects/Publishing/lambda-gateway.ts
```diff
@@ -41,8 +42,8 @@ export class LambdaGateway implements ReleaseEmailGateway {
 
         const req = https.request(options, (res) => {
             if (res.statusCode !== 200) {
-                const error = mapResponseDataToError(res.statusCode, res.statusMessage);
-                throw error;
+                const error = mapResponseDataToError(res.statusCode);
+                reportError(error);
```
DonSeannelly: If we report this new error, don't we loose context on the actual error returned from the lambda function?

PR #342, opened 2019-06-07
Update types and simplify search logic

gshaw1997:
This PR continues the refactor of search by simplifying auth logic and updating types for consistency.


src/LearningObjectSearch/Interactor.ts
```diff
@@ -123,115 +117,65 @@ function formatSearchQuery(
...
+      accessMap[collection] = requestedStatuses;
     }
   }
 
   return accessMap;
 }
 
 /**
- * Builds QueryConditions based on requested collections and collectionAccessMap
+ * Validates and returns authorized statuses for privileged users
```
DonSeannelly: What does "validates" mean here? What is it validating?

PR #343, opened 2019-06-07
Index Submissions into Elasticsearch

DonSeannelly:
This PR introduces a new stage to submitting and cancelling a submission in which the submission is indexed (or removed from the index) in Elasticsearch.

Also, as the scope of the Interactors begins to grow in the submission module, I decided to split them out into their own files. I believe this makes it easier to find the particular use case that one is searching for, and that it makes the task of updating a use case easier as it limits the visible context during any development.

Additionally,
- I have removed the `FileManager` from the module as it was not being used
- I added the `@elastic/elasticsearch` package to make requests to elasticsearch a little easier to configure (the syntax is simpler than HTTP and comes with intellisense)


src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
```diff
@@ -0,0 +1,50 @@
...
+   */
+  async publishSubmission(submission: LearningObject): Promise<void> {
+    await this.client.index({
+      index: INDEX_NAME,
+      type: '_doc',
+      body: cleanLearningObject(submission),
+    });
+  }
+
+  async withdrawlSubmission(learningObjectID: string) {
```
nvisal1: Missing JSDoc

src/LearningObjectSubmission/interactors/cancelSubmission.ts
```diff
@@ -0,0 +1,44 @@
...
+  dataStore: SubmissionDataStore;
+  publisher: SubmissionPublisher;
+  learningObjectId: string;
+  userId: string;
+  user: UserToken;
+  emailVerified: boolean;
+}): Promise<void> {
+  const LearningObjectGateway = LearningObjectAdapter.getInstance();
+  const submission = await params.dataStore.fetchRecentSubmission(params.learningObjectId);
+  if (submission && submission.cancelDate) {
```
nvisal1: We should ensure that the submission exists before canceling. This logic checks to see if the submission has been canceled already, but it will continue to cancel the submission if the submission does not exist.

src/LearningObjectSubmission/interactors/submitForReview.ts
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * updateLearningObjectFields makes requests to update fields on the Learning Object that are affected by
+ * the creation of a submission. Additionally, it requests the README file for the Learning Object is
+ * regenerated to ensure that it has up-to-date information when viewed by a reviewer.
+ * @param params.dataStore storage for Learning Object submissions
+ * @param params.user the user requesting a submission be made for this Learning Object
+ * @param params.learningObjectId id of the learning object to search for
+ * @param params.collection name of collection to submit the Learning Object to
+ */
+async function updateLearningObjectFields(params: { dataStore: SubmissionDataStore; user: UserToken; learningObjectId: string; collection: string; }) {
```
nvisal1: Is there a reason for passing in dataStore? It doesn't look like you are using it here.

src/LearningObjects/LearningObjectAdapter.ts
```diff
@@ -37,5 +38,18 @@ export class LearningObjectAdapter {
       id: params.id,
     });
   }
+
+  async updateLearningObject(params: {
```
nvisal1: Missing JSDoc

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
DonSeannelly: This actually inherits JSDoc from the interface it implements. I didn't see anything else worth noting. I could put 
```
/**
 * @inheritdoc
 */
```
to make it more explicit and readable if need be?

src/LearningObjectSubmission/interactors/submitForReview.ts
DonSeannelly: Good catch - it's not used anymore. I've removed it

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
nvisal1: Yeah, I think that would make it more clear.

DonSeannelly:
I am working on a solution for the failing tests. I'd appreciate a review of the rest of the PR, as the test issue is only related to the environment

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
```diff
@@ -0,0 +1,53 @@
...
+import { cleanLearningObject } from '../shared/elasticsearch';
+import { SubmissionPublisher } from './interactors/SubmissionPublisher';
+import { Client } from '@elastic/elasticsearch';
+
+const INDEX_NAME = 'learning-objects';
+
+export class ElasticsearchSubmissionPublisher implements SubmissionPublisher {
+  client: Client;
+  constructor() {
+    this.client = new Client({ node: process.env.ELASTICSEARCH_DOMAIN });
```
gshaw1997: Should we move the accessing of `process` outside of the constructor execution?

src/LearningObjectSubmission/interactors/SubmissionPublisher.ts
```diff
@@ -0,0 +1,15 @@
+import { LearningObject } from '../../shared/entity';
+export interface SubmissionPublisher {
+  /**
```
gshaw1997: I like the practice of documenting the interface :ok_hand:. It gives context to the caller/implementer and enables hints via intellisense. @Cyber4All/backend I think we should start enforcing in PRs.

src/LearningObjectSubmission/interactors/checkFirstSubmission.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @param collection name of collection to search for in submission collection
+ */
+export async function checkFirstSubmission(params: {
+  dataStore: SubmissionDataStore;
+  collection: string;
+  learningObjectId: string;
+  userId: string;
+  emailVerified: boolean;
+}): Promise<boolean> {
+  return await params.dataStore.fetchSubmission(params.collection, params.learningObjectId)
```
gshaw1997: Should we be loading up the entire document here if we're only checking whether or not it is defined. I think we should have a function to check the document's existence  instead as it could be more efficient if document size increases.

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
DonSeannelly: I'm not opposed to this at all, but may I ask why?

src/LearningObjectSubmission/interactors/checkFirstSubmission.ts
DonSeannelly: I don't think I'm aware of a way to check if a document exists in Mongo without loading it. Do you have ideas for an approach on this?

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
DonSeannelly: I actually ended up resolving this via 2506c48

src/LearningObjectSubmission/interactors/checkFirstSubmission.ts
gshaw1997: I was thinking a `findOne` with just a projection of the `_id` since the rest of the document isn't needed.

This source seems to suggest that a `find` + `limit` of 1; is more efficient though https://blog.serverdensity.com/checking-if-a-document-exists-mongodb-slow-findone-vs-find/

PR #346, opened 2019-06-10
Update  fix released query

gshaw1997:
This PR fixes the issue with the query structure for released searches


src/LearningObjectSearch/drivers/LearningObjectDatastore/ElasticSearchLearningObjectDatastore.ts
```diff
@@ -36,7 +36,7 @@ const SEARCHABLE_FIELDS = [
 ];
 
 const ELASTICSEARCH_DOMAIN = process.env.ELASTICSEARCH_DOMAIN;
-const LEARNING_OBJECT_INDEX = 'learning-objects';
+const LEARNING_OBJECT_INDEX = 'released-objects';
```
DonSeannelly: With the discussion of putting submissions and released objects into the same index the PR I just merged (#343) sets the index to `learning-objects`. Currently the production node has `released-objects` as its index from before that decision, we just need to move the documents

```suggestion
const LEARNING_OBJECT_INDEX = 'learning-objects';
```

DonSeannelly: To update the index, I'm thinking of using the solution outlined here https://stackoverflow.com/questions/28626803/how-to-rename-an-index-in-a-cluster

PR #347, opened 2019-06-10
Refactor express initializations

gshaw1997:
This PR  moves Express Server initializations and config to the root of the service to decrease dependency complexity and allow other modules to more easily expose functionality via the express router.


gshaw1997:
> Why did you decide to remove the `ExpressDriver` in favor of placing that config in the `app.ts` instead of modifying the `ExpressDriver`?

Moving to the root simplifies dependency management by reducing the number of places where we have to rely on a module's dependencies to be passed down. This allows us to more easily add or remove a module's functionality as the module is initialized and exposed to the router at the same level. Otherwise updating would require managing dependencies across multiple files; for initializing the module/providing it's dependencies (at the root) and exposing its functionality by attaching its Express Adapter to the app's router (from within the `ExpressDriver`)

DonSeannelly:
> > Why did you decide to remove the `ExpressDriver` in favor of placing that config in the `app.ts` instead of modifying the `ExpressDriver`?
> 
> Moving to the root simplifies dependency management by reducing the number of places where we have to rely on a module's dependencies to be passed down. This allows us to more easily add or remove a module's functionality as the module is initialized and exposed to the router at the same level. Otherwise adding/detaching a module's functionality would require managing dependencies across multiple files; for initializing the module/providing it's dependencies (at the root) and exposing its functionality by attaching its Express Adapter to the app's router (from within the `ExpressDriver`)

Personally, I don't love this - I like the entrypoint to be lean and only a few lines of code to connect pieces. However, I don't have an argument against it.

gshaw1997:
> > > Why did you decide to remove the `ExpressDriver` in favor of placing that config in the `app.ts` instead of modifying the `ExpressDriver`?
> > 
> > 
> > Moving to the root simplifies dependency management by reducing the number of places where we have to rely on a module's dependencies to be passed down. This allows us to more easily add or remove a module's functionality as the module is initialized and exposed to the router at the same level. Otherwise adding/detaching a module's functionality would require managing dependencies across multiple files; for initializing the module/providing it's dependencies (at the root) and exposing its functionality by attaching its Express Adapter to the app's router (from within the `ExpressDriver`)
> 
> Personally, I don't love this - I like the entrypoint to be lean and only a few lines of code to connect pieces. However, I don't have an argument against it.

@davidnso @nvisal1 I'd like to hear your thoughts on this before a final decision is made.

gshaw1997:
After some noodling on the comments from @DonSeannelly and the initial reasoning for this refactor. I've decided to take a slightly different approach for this refactor.

Although it is a bit easier to manage dependencies from one level of the application, I do not believe that this completely addresses the issue of dependency management we have within our services. I also do favor the lean entry point model, and believe we should minimize its scope to just bootstrapping of the application by initializing globally required modules, dependencies, and components.

So, instead of removing the `ExpressDriver` and moving the configuration of the express `app` to the root, I've refactored the `ExpressDriver` to handle the construction of the express `app` which includes attaching express specific middlewares and routers. 

The significant change to note within the `ExpressDriver` is that it is now not responsible for creating and starting the the HTTP server. Instead it returns the configured express `app` and differs that responsibility back to the caller. This allows for other handlers to be attached to the `app` before the server opens up connections to the service, which we be handled within the bootstrapping process at the service's root.  As an example, this would allow us to easily attach a GraphQL server to our existing express `app`.

PR #348, opened 2019-06-10
Route admin and test traffic to LearningObjectSearch module

gshaw1997:
**Purpose of this PR**
 Exposes functionality of the `LearningObjectSearch` module for consumption by `admin` users. This module allows users to utilize ElasticSearch to perform searches across the system.

**Approach**
Because of changes made in the interactor to decouple from Datastore implementations, a conditional check was added in the router to allow traffic instead of hot-swapping Datastores within the interactor depending on the requester's access level.

**Changes**
- Attaches `LearningObjectSearch`'s `ExpressAdapter` to the application's express router to make the module's functionality consumable.
- Adds middleware within `LearningObjectSearch`'s `ExpressAdapter` to only allow `admin` user's to access the module's exposed functionality.

**Merged Smoke testing into this PR**

***Purpose of this PR***
to create a smoke test environment for Learning Object search. Users with different permissions have access to different Learning Objects when searching.

Users|Visitors have access to Released Learning Objects
Curator@collectionName|Reviewer@collectionName have access to Released Learning Objects and Unreleased Learning Objects that were submitted to their collection for a review.
Editor|Admin has access to all Released and Submitted Learning Objects.
The above test cases live in a Node app within Learning Object Service. This Node app creates the index and mapping for learning-objects in the Elasticsearch test node. It then runs through the above test cases and asserts the response. Circle CI will spin up a docker-compose environment that contains the Node app, Learning Object Service, and the Elasticsearch test node.

The Elasticsearch test node accepts the ELASTIC_SEARCH_TEST_NODE_URI env


.circleci/config.yml
```diff
@@ -27,6 +27,23 @@ jobs:
...
+    <<: *defaults
+    docker:
+      - image: circleci/node:8
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Clone Elasticsearch Smoke Test Repository
+          command: >-
+            cd .. && git clone https://${github_username}:${github_password}@github.com/nvisal1/Elasticsearch-Smoke-Test-Environment.git
```
DonSeannelly: What credentials are being used here?

Also, why did you choose to go with the standalone repository for the tests instead of integrating them inside of this repository?

nvisal1: The tests are inside of this repository. This is pulling the docker-compose config from a private repository.

nvisal1:
> The CircleCI config seems to have a step that clones a git repository, but then I'm seeing the tests in this repository. Could I get some clarification on what is happening here?
> 
> Also, the PR description states that we can "turn on" tests at a later time - but with the addition of the environment variable for toggling search traffic to the new search module, isn't that what we are going to be doing now?

The tests live inside of this repository. Circle CI is pulling docker-compose.yml from a private repository.
I updated the PR description. We no longer need to turn on tests at a later time. Everything is ready to merge.

PR #349, opened 2019-06-11
Extend fetch collection meta search

nvisal1:
This PR changes the `fetchCollectionMeta` function to match collection by `abvName` instead of `name`.


src/drivers/MongoDriver.ts
```diff
@@ -2134,7 +2134,7 @@ export class MongoDriver implements DataStore {
     try {
       const meta: any = await this.db
         .collection(COLLECTIONS.LO_COLLECTIONS)
-        .findOne({ name }, <any>{ name: 1, abstracts: 1 });
+        .findOne({ $or: [ { name }, {abvName: name} ] }, <any>{ name: 1, abstracts: 1 });
```
DonSeannelly: I don't know if I like using the same input to match on two different fields. I think I might prefer switching this to only match on abvName so that it is more a lookup based on a unique identifier.

DonSeannelly: Might be a good idea to first confirm that `collectionService.getCollectionMetadata(abvName)` in the client is only called from the collection page

PR #354, opened 2019-06-13
Add typings for FileMetadataDocument

gshaw1997:
This PR introduces an interface for the documents that will be stored in our file metadata datastore, representing information about a file that has been uploaded to CLARK.


src/FileMetadata/typings/index.ts
```diff
@@ -0,0 +1,69 @@
...
+   * Unique identifier of the file document
+   */
+  id: string;
+  /**
+   * Time at which the file metadata was created (When the file was first ever uploaded)
+   * Value is a string representing the milliseconds elapsed since the UNIX epoch
+   */
+  createdDate: string;
+  /**
+   * User inputted information about the contents of the file
```
DonSeannelly: My editor yells at me for anything that is not standard English -what about this? 

```suggestion
   * User supplied information about the contents of the file
```

src/FileMetadata/typings/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Server generated hash of the file's content
+   */
+  ETag: string;
+  /**
+   * Suffix identifier indicating file's type or characteristics of the file's contents
+   */
+  extension: string;
+  /**
+   * The location the file was uploaded to within the file system's hierarchy
```
DonSeannelly: I think that with the nature of the way we’re doing storage we run the risk of not being clear when we’re talking about the file system as the Learning Object understands it and the file system as we have the files placed in actual storage. What about saying this? 

```suggestion
   * The location the file was uploaded to within the Learning Object’s file system hierarchy
```

src/FileMetadata/typings/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Flag indicating whether or not the file can be included in the Learning Object bundle
+   */
+  packageable: boolean;
+  /**
+   * The size of the file in bytes
+   */
+  size: number;
+  /**
+   * Revision number the file copy is stored under in S3
```
DonSeannelly: I think that it might be helpful to articulate an example of this in the documentation. I don’t think that it naturally is clear. However I can’t quite think of a better name or a better way to phrase it.

PR #356, opened 2019-06-14
Send email only when a top-level Learning Object is released

nvisal1:
The purpose of this PR is to patch release Learning Object emails.

`To avoid spamming emails, we should only send an email when a top-level (parent) Learning Object is released.`


src/LearningObjects/Publishing/interactor.ts
```diff
@@ -11,7 +12,8 @@ export interface PublishingDataStore {
 /**
  * If the user is an admin or editor (the only roles that can release a Learning Object),
  * then request the data store marks the Learning Object as released. Otherwise, throw a
- * ResourceError.
+ * ResourceError. An email will on;y be sent to the Learning Object author if the
```
jbarlow96: PR Looks good, but"only" is spelled wrong here.

src/LearningObjects/Publishing/interactor.ts
```diff
@@ -20,14 +22,21 @@ export async function releaseLearningObject({ userToken, dataStore, releasableOb
     releaseEmailGateway: ReleaseEmailGateway,
 }): Promise<void> {
     if (isAdminOrEditor(userToken.accessGroups)) {
-        releaseEmailGateway.invokeReleaseNotification({
-            learningObjectName: releasableObject.name,
-            authorName: releasableObject.author.name,
-            collection: releasableObject.collection,
-            authorEmail: releasableObject.author.email,
-            username: releasableObject.author.username,
+        const parentObjects = await HierarchyAdapter.getInstance().fetchParents({
```
gshaw1997: Are we loading up all of a Learning Object's parents here to check if the Learning Object is a child object? If so, is the a more efficient way to perform this check?

nvisal1: Would you prefer if I just load the parent ids?

src/LearningObjects/Hierarchy/HierarchyInteractor.ts
```diff
@@ -48,6 +48,26 @@ export async function fetchParents(params: {
...
+ * @returns {Promise<boolean>} false if Learning Object has one or more parents
+ */
+export async function isTopLevelLearningObject(params: {
+  dataStore: DataStore;
+  learningObjectID: string;
+}): Promise<boolean> {
+  const { dataStore, learningObjectID } = params;
+  const parentIds = await dataStore.findParentObjectId({
+    childId: learningObjectID,
+  });
```
DonSeannelly: As we discussed, it might be more clear to return the actual id of the parent object as a string primitive here.

src/LearningObjects/Hierarchy/HierarchyInteractor.ts
```diff
@@ -48,6 +48,26 @@ export async function fetchParents(params: {
...
+  learningObjectID: string;
+}): Promise<boolean> {
+  const { dataStore, learningObjectID } = params;
+  const parentId = await dataStore.findParentObjectId({
+    childId: learningObjectID,
+  });
+  if (parentId) {
+    return false;
+  }
+  return true;
```
DonSeannelly: This could be simplified to
```
return parentId === null
```

PR #359, opened 2019-06-18
Create smoke test environment for Learning Object search

nvisal1:
The purpose of this PR is to create a smoke test environment for Learning Object search. Users with different permissions have access to different Learning Objects when searching.

* `Users`|`Visitors` have access to `Released Learning Objects`
* `Curator@collectionName`|`Reviewer@collectionName` have access to `Released Learning Objects` and `Unreleased Learning Objects` that were submitted to their collection for a review. 
* `Editor`|`Admin` has access to all `Released` and `Submitted Learning Objects`. 

The above test cases live in a Node app within `Learning Object Service`. This Node app creates the `index` and `mapping` for learning-objects in the Elasticsearch test node. It then runs through the above test cases and asserts the response. `Circle CI` will spin up a `docker-compose` environment that contains the Node app, `Learning Object Service`, and the Elasticsearch test node. 

The Elasticsearch test node accepts the `ELASTIC_SEARCH_TEST_NODE_URI` env

This PR can be merged now without changing anything in Circle CI. After Elasticsearch features are fully merged, we can update the Circle CI config to run the new tests.


elasticsearch_testing/gateways/learning-object-service-gateway.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+describe('Sends a request to the learning object service search route', () => {
+
+    beforeAll(async () => {
+        await initElasticsearchNode();
+    });
+
+    it('Search as a user/visitor', async () => {
+        expect.assertions(2);
+        const results = await searchLearningObjects('', 'user/visitor');
```
DonSeannelly: Should `user/visitor` perhaps be `null` or even not defined at all since it is not an actual access group?

elasticsearch_testing/gateways/learning-object-service-gateway.spec.js
```diff
@@ -0,0 +1,54 @@
...
+import { expectedVisitorResponse, expectedEditorAdminResponse } from '../elasticsearch-data/elasticsearch-responses';
+import 'dotenv/config';
+
+describe('Sends a request to the learning object service search route', () => {
+
+    beforeAll(async () => {
+        await initElasticsearchNode();
+    });
+
+    it('Search as a user/visitor', async () => {
```
DonSeannelly: What does this test for in search as a user/visitor?

elasticsearch_testing/gateways/learning-object-service-gateway.spec.js
```diff
@@ -0,0 +1,54 @@
+import { searchLearningObjects } from './learning-object-service-gateway';
+import { initElasticsearchNode } from './elasticsearch_gateway';
+import { expectedVisitorResponse, expectedEditorAdminResponse } from '../elasticsearch-data/elasticsearch-responses';
+import 'dotenv/config';
+
+describe('Sends a request to the learning object service search route', () => {
```
DonSeannelly: Starting the describe statement with a word causes it to read oddly. Perhaps something as simple as:

```suggestion
describe('Learning Object search', () => {
```
or just a rewording of what is there now would help.

elasticsearch_testing/gateways/learning-object-service-gateway.spec.js
nvisal1: This checks the response to make sure that only released objects returned.

DonSeannelly: I think we should update the `it` statement to say that - it will make the tests a little more declarative

elasticsearch_testing/Dockerfile
```diff
@@ -1,14 +1,11 @@
...
 RUN mkdir gateways
 COPY elasticsearch-data ./elasticsearch-data
 COPY gateways ./gateways
-RUN ls
 RUN npm install && npm cache clean --force
 
-RUN npm test
\ No newline at end of file
+
+CMD [ "node_modules/.bin/jest", "--verbose", "--coverage", "--detectOpenHandles" ]
```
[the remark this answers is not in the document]
nvisal1: This was updated.

PR #363, opened 2019-06-19
Implement function to fetch Learning Object by id and revision

gshaw1997:
This PR adds a Datastore method to fetch a Learning Object by the Learning Object's id and revision number.

### DIFF
--- a/src/drivers/MongoDriver.ts
+++ b/src/drivers/MongoDriver.ts
@@ -43,6 +43,7 @@ import { reportError } from '../shared/SentryConnector';
 import { LearningObject, LearningOutcome, User } from '../shared/entity';
 import { Submission } from '../LearningObjectSubmission/types/Submission';
 import { MongoConnector } from '../shared/Mongo/MongoConnector';
+import { mapLearningObjectToSummary } from '../shared/functions';
 
 export enum COLLECTIONS {
   USERS = 'users',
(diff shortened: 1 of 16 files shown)


src/drivers/MongoDriver.ts
```diff
@@ -95,6 +97,39 @@ export class MongoDriver implements DataStore {
...
+    id: string;
+    revision: number;
+  }): Promise<LearningObjectSummary> {
+    const doc =
+      (await this.db
+        .collection(COLLECTIONS.RELEASED_LEARNING_OBJECTS)
+        .findOne({ _id: id, revision })) ||
+      (await this.db
+        .collection(COLLECTIONS.LEARNING_OBJECTS)
+        .findOne({ _id: id, revision }));
```
nvisal1: I remember in a previous discussion you had mentioned that `find limit 1` was more performant than `findOne`. Is there a reason you chose `findOne` here?

gshaw1997: Yes, I found this from a [source](https://blog.serverdensity.com/checking-if-a-document-exists-mongodb-slow-findone-vs-find/) that suggested based on it's benchmarks that is was more performant, but it seems as though this is the case only when, as stated [here](https://dba.stackexchange.com/questions/7573/difference-between-mongodbs-find-and-findone-calls), you are checking for a document's existencen and not returning the document.

PR #366, opened 2019-06-19
Implement LearningObjectGateway for FileMetadata module

gshaw1997:
This PR implements a `ModuleLearningObjectGateway` which proxies request through the `LearningObjectAdapter`. This Gateway allows the `FileMetadata` module to fetch data about Learning Objects.

### DIFF
--- a/src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
+++ b/src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
@@ -1,0 +1,50 @@
+import { LearningObjectGateway } from '../../interfaces';
+import { Requester, LearningObjectSummary } from '../../typings';
+import { LearningObjectAdapter } from '../../../LearningObjects/LearningObjectAdapter';
+
+export class ModuleLearningObjectGateway extends LearningObjectGateway {
+  private adapter: LearningObjectAdapter = LearningObjectAdapter.getInstance();
+
+  /**
+   * @inheritdoc
+   *
+   * Proxies `getWorkingLearningObjectSummary` request to LearningObjectAdapter
+   *
+   * @memberof ModuleLearningObjectGateway
+   */
+  getWorkingLearningObjectSummary(params: {
+    requester: Requester;
+    id: string;
+  }): Promise<LearningObjectSummary> {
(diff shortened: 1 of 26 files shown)


src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
```diff
@@ -0,0 +1,50 @@
+import { LearningObjectGateway } from '../../interfaces';
+import { Requester, LearningObjectSummary } from '../../typings';
+import { LearningObjectAdapter } from '../../../LearningObjects/LearningObjectAdapter';
+
+export class ModuleLearningObjectGateway extends LearningObjectGateway {
```
DonSeannelly: Can you elaborate on why you made the design decision to create this proxy?

src/FileMetadata/interfaces/LearningObjectGateway.ts
```diff
@@ -0,0 +1,43 @@
+import { LearningObjectSummary, Requester } from '../typings';
+
+export abstract class LearningObjectGateway {
```
DonSeannelly: Why did you go with an abstract class here?

gshaw1997: The decision to go with an abstract class is discussed here.
https://github.com/Cyber4All/learning-object-service/pull/337#discussion_r290781423

**TLDR;** Pure abstract classes as interfaces are useful because their names can be used as DI tokens.

src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
gshaw1997: The business logic of this module only needs a small subset of the features provided by the `LearningObjectAdapter` and has its needs defined by the `LearningObjectGateway` interface. This abstraction allows us to swap out `LearningObjectGateway` implementations without having to modify calls from our Interactor.

DonSeannelly: I definitely like the pattern of keeping interfaces small and scoped to their use. I can't say I'm satisfied with the names, but also can't put my finger on why. I'll have to think it over some more - for now, I'm happy with this.

PR #367, opened 2019-06-19
Add AuthManager to FileMetadata module

gshaw1997:
This PR adds an `AuthorizationManager` to the `FileMetadata` module to encapsulate a set of reusable functions used to authorize requests within this module

### DIFF
--- a/src/FileMetadata/AuthorizationManager.ts
+++ b/src/FileMetadata/AuthorizationManager.ts
@@ -4,13 +4,7 @@
 
 import { ResourceError, ResourceErrorReason } from '../shared/errors';
 import { Requester } from './typings';
-
-enum AccessGroup {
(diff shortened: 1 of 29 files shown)


src/FileMetadata/AuthorizationManager.ts
```diff
@@ -0,0 +1,114 @@
...
+
+import { ResourceError, ResourceErrorReason } from '../shared/errors';
+import { Requester } from './typings';
+
+enum AccessGroup {
+  ADMIN = 'admin',
+  EDITOR = 'editor',
+  CURATOR = 'curator',
+  REVIEWER = 'reviewer',
+}
```
DonSeannelly: Perhaps this is a type that we would want to share globally?  It seems that we used it in a lot of places

PR #368, opened 2019-06-19
Add fechRevisionId function

gshaw1997:
This PR implements a Datastore function to retrieve the Learning Object revision number a file belongs to.


gshaw1997:
The functionality added in this PR is no longer needed.

PR #384, opened 2019-06-25
Implement Interactor for FileMetadata Module

gshaw1997:
**Purpose of this PR**
This PR implements methods inside of the `FileMetadata` Module's interactor to perform CRUD operations on file metadata.

Completes https://app.clubhouse.io/clarkcan/story/697/allow-filtering-of-all-file-metadata-by-released-and-not-released and https://app.clubhouse.io/clarkcan/story/698/allow-filtering-of-a-specific-file-s-metadata-by-released-and-not-released


src/FileMetadata/index.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+@expressServiceModule({
+  expressRouter: ExpressHttpAdapter.buildRouter(),
+  providers: [
+    { provide: FileMetaDatastore, useClass: MongoFileMetaDatastore },
+    { provide: LearningObjectGateway, useClass: ModuleLearningObjectGateway },
+  ],
+})
+export class FileMetadata extends ExpressServiceModule {
+  static deleteAllFileMetadata = deleteAllFileMeta;
```
nvisal1: Why is this assignment here?

gshaw1997: It exposes the `deleteAllFileMetadata` function declared within the Interactor to be used by other modules.

DonSeannelly: This is an interesting alternative to creating an adapter class that wraps the interactor functions that other modules can use. I like that the source of truth for the module's capabilities is now named and in one place, but I don't know if I like the syntax here. I wonder if there's a way to make this more declarative. Perhaps even adding the functions to be exposed to the decorator some how?

gshaw1997: @DonSeannelly Could you give an example of what this could look like?

DonSeannelly: Not sure the nuances of it, but potentially something like
```
@({
  ...
  expose: [ interactorOne, interactorTwo ]
})
```

Just sort of musing at this point.

gshaw1997: Ah I see, I did have something similar to this done before, but it was changed because typings were lost when exposing them via the decorator. At run time everything would work because the functions would exist on the object, but Typescript's compiler doesn't understand behavioral changes decorators introduce. 

This limitation is also why the Module's extend Service module classes, to provide typings

PR #389, opened 2019-06-26
Add Query Parameter to Filter Change Logs by Date

jbarlow96:
**_Purpose of this PR_**
The change log resource should be able to take a query parameter that can be given a date and return change logs only written before that date.


src/Changelogs/ChangelogDatastore.ts
```diff
@@ -85,6 +85,38 @@ export class ChangelogDataStore {
     return changelogs;
   }
 
+  /**
+   * Get all changelogs that were created before an inputted date for a specified learning object
```
DonSeannelly: I'd suggest staying away from developer lingo as much as possible, such as making change logs one word or using terms like `inputted`. Maybe something like this:

```suggestion
   * Get all change logs for the specified Learning Object that were created before the given date
```

src/drivers/MongoDriver.ts
```diff
@@ -802,6 +802,16 @@ export class MongoDriver implements DataStore {
     });
   }
 
+  async fetchChangelogsBeforeDate(params: {
```
nvisal1: Missing JSDoc

src/Changelogs/ChangelogDatastore.ts
```diff
@@ -85,6 +85,38 @@ export class ChangelogDataStore {
...
+                input: '$logs',
+                as: 'log',
+                cond: { $lt: [ '$$log.date', parseInt(params.date, 10) ] },
+              },
+            },
+          },
+        },
+      ])
+      .toArray();
+    return changelogs;
```
DonSeannelly: This could be shortened to one statement instead of two.

src/Changelogs/ChangelogInteractor.spec.ts
```diff
@@ -111,3 +111,49 @@ describe('getAllChangelogs', () => {
...
+    return expect(getChangelogs({
+      dataStore,
+      learningObjectId: stubs.learningObject.id,
+      user: {...stubs.userToken, accessGroups: ['editor']},
+      userId: stubs.learningObject.author.id,
+      date: '1',
+    }))
+    .resolves.toHaveLength(1);
+  });
+});
```
DonSeannelly: Since it seems to be the same types of tests for admins and editors, I would suggest using a nested describe block. Something like:
```
describe getChangelogsBeforeDate
  describe when I am an admin
    the two cases...
  describe when I am an editor
    the two cases...
```

src/Changelogs/ChangelogInteractor.ts
```diff
@@ -71,30 +71,34 @@ export async function getRecentChangelog(params: {
 }
 
 /**
- * Fetches all change logs for a specific learning object from the data store .
+ * Handles whether it fetches all change logs for a specific learning object from the data store or if it fetches all changelogs that were created before an inputted date for a specified learning object.
```
DonSeannelly: This comment is a little long. It should be put onto multiple lines, and perhaps reworded a bit for clarity. Try to avoid words like `inputted` and ambiguity like `it`.

src/Changelogs/ChangelogInteractor.spec.ts
```diff
@@ -111,3 +111,49 @@ describe('getAllChangelogs', () => {
     .resolves.toHaveLength(1);
   });
 });
+
+describe('getChangelogsBeforeDate', () => {
```
DonSeannelly: Since these tests are really just checking a specific pathway for `getChangelogs` they should be included in its describe block. Consider using nested describe blocks to break it up into logical groups.

jbarlow96:
> As @DonSeannelly said, the `fetchChangelogsBeforeDate` function can be shortened quite a bit; better for readability and maintainability. The `find` operation with a `$lt` operator may be a better solution here.
> 
> https://docs.mongodb.com/manual/reference/method/db.collection.find/
> 
> https://docs.mongodb.com/manual/reference/operator/query/lt/

Correct me if I'm wrong, but I don't think I can use the `find` operation with a document of this structure. 
image

src/Changelogs/ChangelogDatastore.ts
DonSeannelly: Instead of creating a variable called `changelogs` we can just return the result of the `toArray()` operation.

src/Changelogs/ChangelogInteractor.spec.ts
```diff
@@ -80,80 +80,82 @@ describe('createChangelog', () => {
...
-      dataStore,
-      learningObjectId: stubs.learningObject.id,
-      user: {...stubs.userToken, accessGroups: ['editor']},
-      userId: stubs.learningObject.author.id,
-    }))
-    .resolves.toHaveLength(1);
+describe('getChangelogs', () => {
+  describe('Checking the getAllChangelogs pathway', () => {
+    describe('When I am an admin', () => {
+      it('should return all changelogs for a learning object', async () => {
```
DonSeannelly: I would suggest something more like "when I request all change logs" ... "and I am an admin" ...
instead of mentioning the code pathway. Try to think of the strings provided to these functions as composing the requirements you might get from a Business Analyst.

src/Changelogs/ChangelogDatastore.ts
```diff
@@ -73,7 +73,7 @@ export class ChangelogDataStore {
   async fetchAllChangelogs(params: {
     learningObjectId: string,
   }): Promise<ChangeLogDocument[]> {
-    const changelogs = await this.db
+    return await this.db
```
gshaw1997: `async await` not needed here if we are just returning the Promise.
```suggestion
    return  this.db
```

src/Changelogs/ChangelogDatastore.ts
```diff
@@ -82,7 +82,28 @@ export class ChangelogDataStore {
...
-    return changelogs;
+  }
+
+  /**
+   * Get all change logs for the specified Learning Object that were created before the given date
+   *
+   * @param {string} learningObjectId The id of the learning object that the requested changelog belongs to
+   * @param {string} date The date of that the changelog was created
+   */
+  async fetchChangelogsBeforeDate(params: {
```
gshaw1997: `async await` not needed here if we are just returning the Promise.

```suggestion
  fetchChangelogsBeforeDate(params: {
```

src/Changelogs/ChangelogDatastore.ts
```diff
@@ -82,7 +82,28 @@ export class ChangelogDataStore {
...
+    learningObjectId: string,
+    date: string,
+  }): Promise<ChangeLogDocument[]> {
+    return await this.db
```
gshaw1997: `async await` not needed here if we are just returning the Promise.

```suggestion
    return  this.db
```

PR #390, opened 2019-06-26
Refactor service authorization into shared Authorization Manager

gshaw1997:
**Purpose of this PR**
Because this module encapsulates business use cases and resources regarding Learning Objects within the CLARK system, much of the authorization logic for resources owned by the service is the same throughout the service's modules. To increase maintainability and consistency of this logic, the reusable set of authorization functions and unit tests have been consolidated to the shared `AuthorizationManager`.


src/LearningObjects/Hierarchy/HierarchyInteractor.ts
```diff
@@ -24,7 +23,7 @@ export async function fetchParents(params: {
   let hasFullAccess = false;
   let collectionsWithAccess: string[] = [];
   let requestedByAuthor = false;
-  const [collection, author] = await Promise.all([
```
davidnso: why was the "collection" replaced here?

gshaw1997: It was replaced with a "dont care" since it is an unused variable. The return type of the `await Promise.all` operation is an array and is being destructured into two consts. We need the second element of that array to be set to `author`, but don't care about the first element, so we denote that as `_` by common convention meaning "don't care".

But aside from that, you bring up a good point, why are we requesting a resource we're not using???

PR #392, opened 2019-06-27
Write Learning Object File Bundle to Storage

DonSeannelly:
This PR adds a background task to the Publishing Module, where it will bundle a Learning Object's files on release and save that to S3. This will enable us to refactor Library Service to stream the bundle when downloading released Learning Objects instead of bundling on the fly.

The Bundler was added as a ServiceModule by use of the `node-service-module` package (from @gshaw1997). This has gotten me thinking a lot about how to manage dependencies of modules and stubbing out pieces of the system for testing. In order to get unit tests to work here, I had to refactor the bundling logic that was inside of Library Service to more closely split business logic and implementation detail (we had been fetching data streams in the interactor!).

[ch676]


src/LearningObjects/Publishing/ReleaseEmails/lambda-gateway-factory.ts
```diff
@@ -15,8 +15,8 @@ export class LambdaGatewayFactory {
 
     static buildGateway(): ReleaseEmailGateway {
         switch (process.env.NODE_ENV) {
+            case 'test':
```
nvisal1: Why did you remove the line under testing and add a new case?

package.json
```diff
@@ -28,12 +28,15 @@
...
     "@sentry/node": "^4.6.6",
+    "@types/archiver": "^3.0.0",
+    "archiver": "^3.0.0",
     "aws-sdk": "^2.211.0",
     "cookie-parser": "^1.4.3",
     "cors": "^2.8.4",
     "dotenv": "^6.0.0",
     "express": "^4.16.2",
     "express-jwt": "^5.3.0",
+    "get-uri": "^2.0.3",
```
gshaw1997: Is this package still used? Looks like we're just using `https` now.
```suggestion
```

PR #393, opened 2019-06-28
Revisions/add text users learning objects

pzalep1:
This PR is to add text to the query for the route to fetch a users learning objects in the builder and dashboard.


src/drivers/MongoDriver.ts
```diff
@@ -1561,7 +1562,12 @@ export class MongoDriver implements DataStore {
       } else {
         query.status = { $in: params.status };
       }
-
+      if (params.text) {
+        if (!query.$or) {
+          query.$or = [];
+        }
+        query.$or = query.$or.concat([{ description: { $regex: params.text, $options: 'i' } }, { name: { $regex: params.text, $options: 'i'} }]);
```
gshaw1997: If we're truly just adding to an existing array here and not looking to merge two arrays, could we use `.push` over `.concat`?

Sources suggest `.push` is more performant
https://codeburst.io/jsnoob-push-vs-concat-basics-and-performance-comparison-7a4b55242fa9
https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki

```suggestion
        query.$or.push({ description: { $regex: params.text, $options: 'i' } }, { name: { $regex: params.text, $options: 'i'} });
```

gshaw1997:
To reduce commit bloat in the master branch, let's use the squash and merge strategy here.

PR #394, opened 2019-07-01
Route read operations on file metadata to FileMeta module

gshaw1997:
**Purpose of this PR**
Refactors the service to use the File Metadata module for retrieving file metadata when loading Learning Objects, downloading files, and fetching file metadata.


src/drivers/MongoDriver.ts
```diff
@@ -1428,45 +1459,53 @@ export class MongoDriver implements DataStore {
...
-      throw new ResourceError('Learning Object Author not found', ResourceErrorReason.NOT_FOUND);
+        },
+      ])
+      .toArray();
+    if (results && results[0]) {
+      const object = results[0];
+      object.materials.files = object.orderedFiles;
+      delete object.orderedFiles;
+      const author = await this.fetchUser(object.authorID);
+      return this.generateLearningObject(author, object, params.full);
```
nvisal1: Why did you remove the check for the author? What if the author is undefined?

src/drivers/MongoDriver.ts
```diff
@@ -1538,16 +1584,14 @@ export class MongoDriver implements DataStore {
...
-      const author = await this.fetchUser(object[0].authorID);
-      if (author) {
-        return this.generateReleasedLearningObject(author, object[0], params.full);
-      }
-      throw new ResourceError('Learning Object Author not found', ResourceErrorReason.NOT_FOUND);
+    if (results && results[0]) {
+      const object = results[0];
+      object.materials.files = object.orderedFiles;
+      delete object.orderedFiles;
+      const author = await this.fetchUser(object.authorID);
```
nvisal1: Why did you remove the check for the author? What if the author is undefined?

src/drivers/MongoDriver.ts
gshaw1997: The author should never be undefined if the Learning Object exists, if it is, there's data inconsistency within the system, and is not a result of the requester requesting a resource that doesn't exist. I would say at that point its an Internal Service Error.

Additionally, I believe we should continue to handle throwing these sorts of errors at the interactor level, since it knows the full context of the request and can convey errors in relation to that. In this case, I think it's confusing for a client requesting a Learning Object to receive an error about a user resource not existing.

src/drivers/MongoDriver.ts
gshaw1997: The author should never be undefined if the Learning Object exists, if it is, there's data inconsistency within the system, and is not a result of the requester requesting a resource that doesn't exist. I would say at that point its an Internal Service Error.

Additionally, I believe we should continue to handle throwing these sorts of errors at the interactor level, since it knows the full context of the request and can convey errors in relation to that. In this case, I think it's confusing for a client requesting a Learning Object to receive an error about a user resource not existing.

PR #396, opened 2019-07-01
Refactor File Operations to a Module

nvisal1:
***Purpose of this PR***
* Move single file download logic to the `FileManager` module
* Expose functions for deleting files and folders in the `FileManager` module
* Refactor `FileManager` module to follow the Service Module pattern
* Remove legacy file upload code
* Remove the `URL` property from Learning Object PDF -> `breaking change`


src/FileManager/adapters/FileManagerAdapter.ts
```diff
@@ -1,15 +1,11 @@
-import { FileManager } from '../shared/interfaces/interfaces';
-import { uploadFile } from './FileInteractor';
-import { FileUpload } from '../shared/interfaces/FileManager';
-export class FileManagerAdapter {
+import { uploadFile } from '../Interactor';
+import { FileUpload } from '../typings/file-manager';
+import { FileManagerAdapter as FileManagerAdapterInterface } from '../interfaces/FileManagerAdapter';
+export class FileManagerAdapter implements FileManagerAdapterInterface {
```
gshaw1997: If using the `ServiceModule` could we export the functions we want to expose from this module like done [here](https://github.com/Cyber4All/learning-object-service/pull/384#discussion_r297806866), instead for consistency. I know this isn't documented anywhere yet, but will do so when going through and writing up comprehensive docs for using the pattern.

src/FileManager/adapters/FileManagerAdapterStub.ts
```diff
@@ -0,0 +1,22 @@
+
```
gshaw1997: What's the purpose of this stub? Shouldn't the dependent of this dependency provide this mock for tests?

src/FileManager/typings/file-manager.ts
```diff
@@ -0,0 +1,54 @@
+import { CompletedPartList, CompletedPart } from 'aws-sdk/clients/s3';
+
+// Export aliased types for ease of update in the case that AWS S3 is no longer the driver used.
+export type CompletedPart = CompletedPart;
+export type CompletedPartList = CompletedPartList;
+
+export interface FileUpload {
```
gshaw1997: Is this still needed?

src/FileManager/typings/file-manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    path: string;
+    // FIXME: This should define the specific types it can take
+    data: any;
+  }
+
+export interface MultipartFileUpload extends FileUpload {
```
gshaw1997: Is this still needed?

src/FileManager/typings/file-manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    partNumber: number;
+    uploadId: string;
+}
+
+export interface MultipartFileUploadStatus {
```
gshaw1997: Is this still needed?

src/FileManager/typings/file-manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    _id: string;
+    path: string;
+    completedParts: CompletedPartList;
+    createdAt: string;
+}
+export interface MultipartFileUploadStatusUpdates {
```
gshaw1997: Is this still needed?

src/FileManager/typings/file-manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    partsUploaded: number;
+    bytesUploaded: number;
+}
+
+export interface MultipartUploadData {
```
gshaw1997: Is this still needed?

src/FileManager/typings/file-manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    uploadId?: string;
+    url?: string;
+    completedPart?: CompletedPart;
+}
+
+export interface DZFileMetadata {
```
gshaw1997: Is this still needed?

src/FileManager/typings/file-manager.ts
```diff
@@ -0,0 +1,54 @@
...
+    dzchunkindex?: number;
+    dztotalfilesize?: number;
+    dzchunksize?: number;
+    size?: number;
+    dztotalchunkcount?: number;
+    dzchunkbyteoffset?: number;
+    fullPath: string;
+}
+
+export interface DZFile extends DZFileMetadata {
```
gshaw1997: Is this still needed?

src/FileMetadata/adapters/FileMetadataAdapter.ts
```diff
@@ -0,0 +1,39 @@
+import { LearningObjectFile, FileMetadataFilter } from '../typings';
```
gshaw1997: An adapter isn't needed here. The functions are already exposed via the module

src/LearningObjects/PDFKitDriver.ts
```diff
@@ -1,9 +1,9 @@
 import * as PDFKit from 'pdfkit';
 import { LEARNING_OBJECT_ROUTES } from '../shared/routes';
 import { FileManager } from '../shared/interfaces/interfaces';
-import { FileUpload } from '../shared/interfaces/FileManager';
 import * as striptags from 'striptags';
 import { LearningObject, LearningOutcome } from '../shared/entity';
+import { FileUpload } from '../FileManager/typings/file-manager';
```
gshaw1997: What is this used for?

src/LearningObjects/LearningObjectRouteHandler.spec.ts
```diff
@@ -48,7 +48,7 @@ describe('LearningObjectRouteHandler', () => {
         dataStore = await MongoDriver.build(global['__MONGO_URI__']);
         HierarchyAdapter.open(dataStore);
         fileManager = new MockS3Driver();
-        FileManagerAdapter.open(fileManager);
+        FileManagerAdapter.open();
```
gshaw1997: If using the `ServiceModule` we'll want to initialize the module, but shouldn't this dependency be stubbed/mocked?

src/app.ts
```diff
@@ -85,7 +85,7 @@ function initModules() {
   HierarchyAdapter.open(dataStore);
   LearningObjectAdapter.open(dataStore, fileManager, library);
   LearningObjectSearch.initialize();
-  FileManagerAdapter.open(fileManager);
+  FileManagerAdapter.open();
```
gshaw1997: If using the `ServiceModule` we'll want to initialize here.

src/drivers/MongoDriver.ts
```diff
@@ -1933,6 +1821,39 @@ export class MongoDriver implements DataStore {
     return query;
   }
 
+  async findSingleFile(params: {
```
gshaw1997: Is this still needed in the MongoDriver?

src/shared/interfaces/interfaces.ts
```diff
@@ -1,6 +1,6 @@
 
 import { DataStore } from './DataStore';
-import { FileManager } from './FileManager';
+import { FileManager } from '../../FileManager/interfaces/FileManager';
```
gshaw1997: Do we need this interface exposed at the global service level?

src/tests/mock-drivers/MockS3Driver.ts
```diff
@@ -1,7 +1,7 @@
 import { FileManager } from '../../shared/interfaces/interfaces';
-import { FileUpload, CompletedPartList } from '../../shared/interfaces/FileManager';
 import { CompletedPart } from 'aws-sdk/clients/s3';
 import { Readable } from 'stream';
+import { FileUpload, CompletedPartList } from '../../FileManager/typings/file-manager';
```
gshaw1997: Is this still needed?

src/shared/interfaces/DataStore.ts
```diff
@@ -206,12 +191,10 @@ export interface DataStore
     fileId: string;
     description: string;
   }): Promise<LearningObject.Material.File>;
-
-  // File Uploads
-  updateMultipartUploadStatus(params: {
-    id: string;
-    completedPart: CompletedPart;
-  }): Promise<void>;
+  findSingleFile(params: {
```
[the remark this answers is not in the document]
nvisal1: I decided to keep this function because it is used by the `updateFileDescription` function in the `MongoDriver` and the `removeFile` function in `LearningObjects/LearningObjectInteractor`.

Can you confirm if these functions are deprecated?

src/drivers/drivers.ts
```diff
@@ -3,7 +3,7 @@ import { ExpressRouteDriver } from './express/ExpressRouteDriver';
 import { ExpressAuthRouteDriver } from './express/ExpressAuthRouteDriver';
 import { ExpressAdminRouteDriver } from './express/ExpressAdminRouteDriver';
 import { MongoDriver } from './MongoDriver';
-import { S3Driver } from '../FileManager/S3Driver';
+import { S3Driver } from '../FileManager/drivers/FileManagerDriver/S3Driver';
```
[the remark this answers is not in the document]
nvisal1: It is being used in the `LearningObjects` module in the `updateReadme` function.

src/FileManager/Interactor.ts
```diff
@@ -54,7 +104,7 @@ export async function downloadSingleFile(params: {
   }
 
   if (!params.filter || params.filter === 'released') {
-    fileMetaData = await FileMetadata.geFileMetadata({
+    fileMetaData = await FileMetadata.getFileMetadata({
```
gshaw1997: We should swap out direct call to the module with a call to an interface that uses this module, like a module gateway

src/FileManager/Interactor.ts
```diff
@@ -70,7 +120,7 @@ export async function downloadSingleFile(params: {
 
     // To maintain existing functionality, service token is elevated to author to fetch unreleased file meta
     serviceToken.username = learningObject.author.username;
-    fileMetaData = await FileMetadata.geFileMetadata({
+    fileMetaData = await FileMetadata.getFileMetadata({
```
gshaw1997: We should swap out direct call to the module with a call to an interface that uses this module, like a module gateway

src/FileManager/Interactor.ts
```diff
@@ -1,11 +1,18 @@
+import { FileManagerModule as Module } from '.';
 import { DataStore } from '../shared/interfaces/DataStore';
```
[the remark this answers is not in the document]
nvisal1: It is used in the `Drivers` namespace to resolve module dependencies.

src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
```diff
@@ -5,6 +5,8 @@ import {
   LearningObjectFile,
 } from '../../typings';
 import { LearningObjectAdapter } from '../../../LearningObjects/LearningObjectAdapter';
+import { UserToken } from '../../../shared/types';
```
gshaw1997: Can we move this to a typings internal to the module?

src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
```diff
@@ -5,6 +5,8 @@ import {
...
+import { LearningObject } from '../../../shared/entity';
```
gshaw1997: Can we move this to a typings internal to the module?

src/FileManager/adapters/FileManagerAdapter.ts
nvisal1: Yeah, I totally missed that thread.

src/shared/interfaces/DataStore.ts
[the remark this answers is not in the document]
gshaw1997: Ah I see, yes they are now deprecated :ok_hand:

src/drivers/drivers.ts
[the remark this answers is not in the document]
gshaw1997: Could we swap this for a FileGateway that the `updateReadme` function could use?

src/FileManager/Interactor.ts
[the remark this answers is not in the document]
gshaw1997: Ah I see, idk how I missed that haha. Thanks for clarifying.

src/FileManager/adapters/FileManagerAdapterStub.ts
nvisal1: I'm not certain. I believe it was authored by @DonSeannelly.

src/FileManager/typings/file-manager.ts
nvisal1: This is still used in many places.

src/LearningObjects/PDFKitDriver.ts
nvisal1: This function is used by the `function addEventListeners` function.

src/LearningObjects/LearningObjectRouteHandler.spec.ts
nvisal1: The FileManagerAdapter was deleted in favor of exposing static functions at the module index.

src/app.ts
nvisal1: The FileManagerAdapter was deleted in favor of exposing static functions at the module index.

src/drivers/MongoDriver.ts
nvisal1: I decided to keep this function because it is used by the updateFileDescription function in the MongoDriver and the removeFile function in LearningObjects/LearningObjectInteractor.

Can you confirm if these functions are deprecated?

src/tests/mock-drivers/MockS3Driver.ts
nvisal1: FileUpload will still be needed. I think this that file should be moved into the `FileManager` module. Thoughts?

src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
nvisal1: Should `UserToken` be reimplemented in every module that uses it?

src/FileMetadata/drivers/LearningObjectGateway/ModuleLearningObjectGateway.ts
nvisal1: Should `Entity` types be redefined in every module that uses them?

src/drivers/MongoDriver.ts
gshaw1997: They are indeed deprecated

src/tests/mock-drivers/MockS3Driver.ts
gshaw1997: I would agree, we should move all file ops to the `FileManager`

src/shared/interfaces/interfaces.ts
nvisal1: Mock S3 driver should be a stub gateway that lives in the Learning Object module

src/FileManager/Interactor.ts
```diff
@@ -0,0 +1,261 @@
...
+  authorUsername: string;
+  learningObjectId: string;
+  path: string;
+}): Promise<void> {
+  await Drivers.fileManager().deleteFolder({
+    authorUsername,
+    learningObjectId,
+    path,
+  });
+}
```
DonSeannelly: A few of these interactor functions just proxy requests to the fileManager. What was the reason for this, as compared to giving other modules access to inject this driver?

src/FileManager/Interactor.ts
```diff
@@ -0,0 +1,261 @@
...
+
+/**
+ * Sends a file back to the caller as a readable stream.
+ *
```
DonSeannelly: There's quite a bit of complexity here in this function. We should update the doc to note the nuances.

src/FileManager/drivers/FileMetadataGateway/FileMetadataModuleGateway.ts
```diff
@@ -0,0 +1,24 @@
...
+     */
+    getFileMetadata(params: {
+        requester: UserToken;
+        learningObjectId: string;
+        id: string;
+        filter: FileMetadataFilter;
+    }): Promise<LearningObject.Material.File> {
+        return FileMetadataModule.getFileMetadata(params);
+    }
+}
```
DonSeannelly: What is the driving factor behind the `ModuleGateway` proxy pattern?

Just sort of thinking here, but why not just have the module declare an interface that it expects, and then drop the `FileMetadataModule` in?

src/LearningObjects/Publishing/Bundler/Interactor.ts
```diff
@@ -96,21 +95,18 @@ function addCCLicense(prefix: string = ''): BundleData {
...
-  uri,
   name,
   prefix = '',
 }: {
-  uri: string;
   name: string;
   prefix?: string;
 }): BundleData {
-  return { name, prefix, uri };
+  return { name, prefix, uri: null };
```
DonSeannelly: Is this permanent, or should we have a FIXME/TODO?

src/LearningObjects/Publishing/interactor.ts
```diff
@@ -25,22 +29,32 @@ export interface PublishingDataStore {
...
-    }
-    await dataStore.addToReleased(releasableObject);
-    await sendEmail(releasableObject, userToken, releaseEmailGateway);
+  if (!requesterIsAdminOrEditor(userToken)) {
+    throw new ResourceError(
+      `${
+        userToken.username
+      } does not have access to release this Learning Object`,
+      ResourceErrorReason.INVALID_ACCESS,
+    );
```
DonSeannelly: @gshaw1997 Can we update local prettier config to match the repo's tslint rules (or vice versa) so that we can avoid all of these diffs

src/LearningObjects/Publishing/interactor.ts
```diff
@@ -52,25 +66,22 @@ export async function releaseLearningObject({ userToken, dataStore, releasableOb
...
  * @param userToken the user who has requested to publish a Learning Object
  */
-async function createPublishingArtifacts(releasableObject: LearningObject, userToken: UserToken) {
-    const storagePrefix = `${releasableObject.author.username}/${releasableObject.id}`;
-    FileManagerAdapter.getInstance().uploadFile({
-        file: {
-            path: `${storagePrefix}/meta.json`,
-            data: JSON.stringify(releasableObject.toPlainObject()),
-        },
-    });
```
DonSeannelly: @gshaw1997 It looks like we removed creation of the `meta.json`. Was this an accident?

src/drivers/MongoDriver.ts
```diff
@@ -1832,64 +1687,6 @@ export class MongoDriver implements DataStore {
...
-      );
-      return this.findSingleFile({
-        learningObjectId: params.learningObjectId,
-        fileId: params.fileId,
-      });
-    } catch (e) {
-      return Promise.reject(e);
-    }
-  }
-
```
DonSeannelly: image

src/shared/functions/index.ts
```diff
@@ -130,3 +130,30 @@ export function mapAuthorToSummary(author: Partial<User>): AuthorSummary {
     organization: author.organization,
   };
 }
+
+/**
+ * Title cases string
```
DonSeannelly: This JSDoc description says nothing more than the function name does. Small nitpick here, but maybe something like: 
```suggestion
 * Capitalizes the first letter of each word in the provided string.
```

src/shared/types/file-upload.ts
```diff
@@ -0,0 +1,7 @@
+import { Readable } from 'stream';
+
+export interface FileUpload {
+    path: string;
+    // FIXME: This should define the specific types it can take
+    data: Buffer|Uint8Array|Blob|string|Readable;
```
DonSeannelly: It looks like this FIXME may have been resolved? I believe I put it here when `data` had a type of `any`.

src/FileManager/drivers/FileMetadataGateway/FileMetadataModuleGateway.ts
gshaw1997: It was supposed to serve as an abstraction layer, but you make a good point. If we declare the interface the interactor is expecting, we'll be able to drop in the module directly as the dependency, as long as those two interfaces are consistent. What do we do in the case where they are not consistent? Do would we instead create this `ModuleGateway` (or at that point `Adapter` may be a more appropriate name)? If so, why not just do that from the start, to maintain consistency?

src/LearningObjects/Publishing/Bundler/Interactor.ts
gshaw1997: This should have a FIXME, but it is fixed in this PR https://github.com/Cyber4All/learning-object-service/pull/404 since pdf urls were removed.

src/LearningObjects/Publishing/interactor.ts
gshaw1997: Yup, was definitely an accident. Good catch!

src/FileManager/Interactor.ts
gshaw1997: We wanted to move all business logic related to file management to a single place and have the `FileManager` module own that domain. Although some functions just proxy the `fileManager` driver, there are operations that require logic.

src/FileManager/drivers/FileMetadataGateway/FileMetadataModuleGateway.ts
DonSeannelly: Personally, I'm a fan of keeping it as simple as possible and only scaling up the patterns in use when needed. I haven't yet seen us actually need to use an Adapter in this way, so I would say if we need to in the future we can easily set that up, but in the other cases where it is a direct mapping we can keep it simpler.

src/LearningObjects/Publishing/interactor.ts
gshaw1997: Is there a good way to sync the two? Most options don't seem to have a one-to-one mapping from ts-lint config and prettier config

tslint.json
```diff
@@ -11,7 +11,7 @@
     "interface-name": [false],
     "jsdoc-format": true,
     "label-position": true,
-    "max-line-length": [true, 180],
+    "max-line-length": [true, 80],
```
DonSeannelly: @gshaw1997 What’s the reasoning for choosing the number 80?

.prettierrc
```diff
@@ -0,0 +1,14 @@
...
+  "htmlWhitespaceSensitivity": "css",
+  "jsxBracketSameLine": true,
+  "jsxSingleQuote": true,
+  "printWidth": 80,
+  "proseWrap": "preserve",
+  "quoteProps": "as-needed",
+  "semi": true,
+  "tabWidth": 2,
+  "trailingComma": "es5"
+}
```
DonSeannelly: This commit might be better off going in its own PR. How do we plan to include prettier in each developers workflow? I know there’s a VS code extension and a CLI. How can we make this consistent for developers? I’m thinking we might need to update the read me with some extra steps for getting started developing in this repository

tslint.json
gshaw1997: This decision was based on the recommendation by Prettier https://prettier.io/docs/en/options.html#print-width

src/FileManager/Interactor.ts
DonSeannelly: I think that I’m okay with this for now, but the question that’s coming up in my mind is perhaps we need to identify a way to break up business logic when it’s not quite an interactor by the traditional sense

PR #399, opened 2019-07-01
Davidns/ch696/allow filtering of a user s learning object

davidnso:
This PR handles the following cases for searching and filtering a user's learning objects. 

- If the requester is not the author and is not a privileged user, only summaries of released Learning Objects should be returned. The not released filter cannot be applied in this case.

-   If the requester is the author, and has not specified a filter, summaries of released and all not released Learning Object should be returned.

    If they have specified filtering by released, only summaries of released Learning Objects should be returned.
    If they have specified filtering by not released, only summaries of not released Learning Objects should be returned.

 -   If the requester is a curator/reviewer of a collection/collections of the user's Learning Objects and has not specified a filter, summaries of released and review stage Learning Objects in their collection should be returned.
   If they have specified filtering by released, only summaries of released Learning Objects should be returned.
    If they have specified filtering by not released, only summaries of review stage in their collection Learning Objects should be returned.

-    If the requester is a admin/editor and has not specified a filter, summaries of released and all review stage Learning Objects should be returned.
    If they have specified filtering by released, only summaries of released Learning Objects should be returned.
    If they have specified filtering by not released, only summaries of review stage Learning Objects should be returned.

side note: **I'm having an issue with booleans in my dev environment so any objects that seem like they should be booleans but are actually strings are a result of that.**


src/drivers/express/ExpressRouteDriver.ts
```diff
@@ -107,19 +107,20 @@ export class ExpressRouteDriver {
       try {
         const query = req.query;
         const userToken: UserToken = req.user;
+        const released: string = query.released;
```
gshaw1997: How does a user specify another status? Or multiple statuses?

src/interactors/LearningObjectInteractor.ts
```diff
@@ -164,6 +164,107 @@ export class LearningObjectInteractor {
...
     }
   }
+  /**
+   * searches for a users learning objects, optionally applying a filter by status or bundled status's
+   * and performs authorization based on the requesters
+   * access privileges
+   * @async
+   *
+   * @returns {LearningObjectSummary[]} the user's learning objects found by the query
+   * @param params.dataStore
```
gshaw1997: Add descriptions to params

src/interactors/LearningObjectInteractor.ts
```diff
@@ -164,6 +164,107 @@ export class LearningObjectInteractor {
...
+        text,
+        orderBy,
+        sortType,
+        collection,
+        status,
+      } = this.formatSearchQuery(query);
+
+      let response: LearningObjectSummary[];
+
+      await dataStore.findUser(username);
```
gshaw1997: What's the purpose of this?

src/interactors/LearningObjectInteractor.ts
```diff
@@ -164,6 +164,107 @@ export class LearningObjectInteractor {
...
+        response = await dataStore.searchReleasedUserObjects(
+            query,
+            username,
+        );
+      }
+
+      return response;
+
+    } catch (err) {
+      console.log(err);
```
gshaw1997: Error logs not allowed

src/interactors/LearningObjectInteractor.ts
```diff
@@ -1117,6 +1218,17 @@ export class LearningObjectInteractor {
     return status;
   }
 
+  private static getRequestedStatusFilter(released: string, isAuthor?: boolean): string[] {
+    switch (released) {
+      case 'true': return LearningObjectState.RELEASED; break;
```
gshaw1997: `break` not needed if returning.

src/interactors/LearningObjectInteractor.ts
```diff
@@ -1117,6 +1218,17 @@ export class LearningObjectInteractor {
     return status;
   }
 
+  private static getRequestedStatusFilter(released: string, isAuthor?: boolean): string[] {
```
gshaw1997: Not entirely following the purpose of this here.

src/shared/functions/index.ts
```diff
@@ -130,3 +131,10 @@ export function mapAuthorToSummary(author: Partial<User>): AuthorSummary {
     organization: author.organization,
   };
 }
+
+export function mapChildToSummary(child: Partial<LearningObject>): LearningObjectChildSummary {
```
gshaw1997: JSDoc

src/shared/interfaces/DataStore.ts
```diff
@@ -127,6 +127,16 @@ export interface DataStore
     total: number;
     objects: LearningObject[];
   }>;
+
+  searchReleasedUserObjects(
```
gshaw1997: JSDoc

src/shared/interfaces/DataStore.ts
```diff
@@ -127,6 +127,16 @@ export interface DataStore
...
+    query: LearningObjectQuery,
+    username: string,
+  ): Promise<LearningObjectSummary[]>;
+  searchAllUserObjects(
```
gshaw1997: JSDoc

src/shared/types/learning-object-summary.ts
```diff
@@ -42,6 +53,10 @@ export interface LearningObjectSummary {
    * The collection the Learning Object was submitted to
    */
   collection: string;
+  /**
+   * Summary of Learning object child objects
+   */
+  children?: LearningObjectChildSummary[];
```
gshaw1997: This property should always be defined. If the object has no children, it will just be defined as an  empty array
```suggestion
  children: LearningObjectChildSummary[];
```

src/interactors/LearningObjectInteractor.ts
```diff
@@ -164,6 +164,107 @@ export class LearningObjectInteractor {
...
+      if (userToken) {
+
+        const isAuthor = this.hasReadAccess({
+          userToken,
+          resourceVal: params.username,
+          authFunction: checkAuthByUsername,
+        });
+        if (isAuthor) {
+          status = status ? status : this.getRequestedStatusFilter(released, true);
+          console.log(status);
```
gshaw1997: No logs
```suggestion
```

src/interactors/LearningObjectInteractor.ts
```diff
@@ -164,6 +164,107 @@ export class LearningObjectInteractor {
...
+   * @returns {LearningObjectSummary[]} the user's learning objects found by the query
+   * @param params.dataStore
+   * @param params.library
+   * @param params.username
+   * @param params.userToken
+   * @param params.loadChildren
+   * @param params.query
+   */
+
+  public static async searchUsersObjects(params: {
```
gshaw1997: Complexity of this function is extremely high,  making it difficult to read through. Lets pair up on this one and see if we can simplify.

src/drivers/express/ExpressRouteDriver.ts
davidnso: the specified status's come in through the `query`, as `query.status` since it's still a `learningobjectquery`

src/interactors/LearningObjectInteractor.ts
davidnso: It'll throw an error if the username requested is not found

src/interactors/LearningObjectInteractor.ts
davidnso: just to clean up certain things going on in the interactor function instead of having to repeat the same lines of code for the difference use cases. For the different access groups, whether they are an author, admin/editor, reviewer or curator we need to check whether or not they requested a general status filter and return the requested status's based on either the general status filter or a specific status. In the current implementation, the specified status's take priority. so in the niche case where a user asks for `not-released` and `status=waiting` they will only see objects with a status of waiting.

src/interactors/LearningObjectInteractor.ts
davidnso: Yea it's a bit of a beast and definitely needs to be simplified if possible.

src/shared/types/learning-object-summary.ts
davidnso: Yea, the map function in the driver handles that

src/shared/functions/index.ts
```diff
@@ -91,6 +95,22 @@ export function toNumber(value: any): number {
...
+ */
+export function toBoolean(value: any): boolean {
+  if (value === 'true' || value === true) {
+    return true;
+  }
+  if (value === 'false' || value === false) {
+    return false;
+  }
+  return false;
+}
```
DonSeannelly: What is the reason for using this function over just a loose equality check?

davidnso: For the sake of keeping things consistent when formatting the search query I suppose ensuring that all the values are what we expect them to be, making the query object given to the interactor easier to work with

gshaw1997: @DonSeannelly How would a loose equality check work here? `true=='true' | false=='false'` are both false.

DonSeannelly: I must not have had my coffee yet... @gshaw1997 you're totally right

src/drivers/MongoDriver.ts
```diff
@@ -347,6 +348,100 @@ export class MongoDriver implements DataStore {
...
+   *
+   * @param {LearningObjectQuery} query query containing  for field searching
+   * @param {string} username username of an author in CLARK
+   */
+  async searchReleasedUserObjects(
+    query: ReleasedUserLearningObjectSearchQuery,
+    username: string,
+  ): Promise<LearningObjectSummary[]> {
+    const { text } = query;
+    const authorID = await this.findUser(username);
```
nvisal1: This function is asking for a username, but needs a userId. I think it makes more sense to fetch the userId at the interactor level.

src/drivers/MongoDriver.ts
```diff
@@ -347,6 +348,100 @@ export class MongoDriver implements DataStore {
...
+    const searchQuery: any = {
```
nvisal1: We should set a type for this and avoid `any`.

src/drivers/MongoDriver.ts
```diff
@@ -347,6 +348,100 @@ export class MongoDriver implements DataStore {
...
+
+    let orConditions: QueryCondition[] = [];
+    if (collectionRestrictions) {
+      const conditions: QueryCondition[] = this.buildCollectionQueryConditions(
+        collectionRestrictions,
+      );
+      orConditions = this.buildQueryConditions(conditions);
+    }
+
+    const searchQuery: any = {
```
nvisal1: We should set a type for this and avoid any.

src/drivers/MongoDriver.ts
```diff
@@ -347,6 +348,100 @@ export class MongoDriver implements DataStore {
...
+      authorID,
+    };
+    if (revision != null) {
```
nvisal1: This could be rewritten as `if (revision)`

src/drivers/MongoDriver.ts
```diff
@@ -347,6 +348,100 @@ export class MongoDriver implements DataStore {
...
+      resultSet.map(learningObject => {
+        return this.generateReleasedLearningObjectSummary(learningObject);
+      }),
+    );
+  }
+
+  /**
+   * Performs aggregation to join the users objects from the released and working collection before
+   * searching and filtering
+   *
```
nvisal1: This function seems pretty complex. Please provide more in depth docs

src/drivers/MongoDriver.ts
nvisal1: Also, the findUser function is deprecated. Why are we using it here?

src/drivers/MongoDriver.ts
```diff
@@ -1355,6 +1453,30 @@ export class MongoDriver implements DataStore {
...
 
+  /**
+   * @inheritdoc
+   * @async
+   *
+   * @param {string} username the user's username or email
+   *
+   * @returns {UserID}
+   */
+  async findUserId(username: string): Promise<string> {
```
nvisal1: Since the username can be a username or email, we should change the name of this argument.

src/drivers/MongoDriver.ts
```diff
@@ -2473,17 +2599,65 @@ export class MongoDriver implements DataStore {
...
+      status = status ? status : [LearningObject.Status.RELEASED];
+      children = await this.loadChildObjects({
+        id: record._id,
+        status,
+        full: false,
+      });
+    }
+
+    return mapLearningObjectToSummary({
+      ...(record as any),
```
nvisal1: Try to avoid the `any` type.

src/drivers/MongoDriver.ts
```diff
@@ -347,6 +348,100 @@ export class MongoDriver implements DataStore {
...
+   * @param {UserLearningObjectSearchQuery} query query containing status and text for field searching
+   * @param {string} username username of an author in CLARK
+   * @param {QueryCondition} conditions Array containing a reviewer or curators requested collections.
+   *
+   * @returns {LearningObjectSummary[]}
+   * @memberof MongoDriver
+   */
+  async searchAllUserObjects(
+    query: UserLearningObjectSearchQuery,
+    username: string,
```
nvisal1: This function is asking for a username, but needs a userId. I think it makes more sense to fetch the userId at the interactor level.

nvisal1: FindUser is deprecated

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 We decided to do this at the driver level because the datastore, the owner of the data, has full context on how the data is stored, so it should be responsible of fetching information about foreign keys needed to fulfill the specified request, in this case, fetching summaries of a user's Learning Objects.

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 We decided to do this at the driver level because the datastore, the owner of the data, has full context on how the data is stored, so it should be responsible of fetching information about foreign keys needed to fulfill the specified request, in this case, fetching summaries of a user's Learning Objects.

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 `revision` is a positive integer from 0-n, Shortening to `if(revision)` would result in this block not being called if `revision=0`

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 Any suggestions to make this more clear?

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 We decided to do this at the driver level because the datastore, the owner of the data, has full context on how the data is stored, so it should be responsible of fetching information about foreign keys needed to fulfill the specified request, in this case, fetching summaries of a user's Learning Objects.

src/drivers/MongoDriver.ts
gshaw1997:
```suggestion
    const searchQuery: {[index: string]: any} = {
```

src/drivers/MongoDriver.ts
nvisal1: @gshaw1997 maybe something like `userIdentifier`. A more general name to convey that it accepts two identifiers.

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 `any` is used because the shape of this query is not definitive and is subject to change. I do think we can be a bit more specific here and allow flexibility in its shape using the suggested update below.

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 any is used because the shape of this query is not definitive and is subject to change. I do think we can be a bit more specific here and allow flexibility in its shape using the suggested update below.

gshaw1997:
```suggestion
    const searchQuery: {[index: string]: any} = {
```

src/drivers/MongoDriver.ts
gshaw1997: @nvisal1 any is used to allow compatible properties on `record` to be passed to the `mapLearningObjectToSummary` function to avoid needing to constantly update the function if we want to expose more stored properties within the `LearningObjectSummary` type

src/drivers/MongoDriver.ts
```diff
@@ -347,6 +348,100 @@ export class MongoDriver implements DataStore {
...
+    collectionRestrictions?: CollectionAccessMap,
+  ): Promise<LearningObjectSummary[]> {
+    const { revision, status, text } = query;
+    const authorID = await this.findUser(username);
```
gshaw1997:
```suggestion
    const authorID = await this.findUserId(username);
```

src/drivers/MongoDriver.ts
gshaw1997:
```suggestion
    const authorID = await this.findUserId(username);
```

src/shared/interfaces/DataStore.ts
```diff
@@ -171,6 +204,14 @@ export interface DataStore
   }): Promise<MultipartFileUploadStatus>;
 
   // Users
+  /**
+   *
+   * lookup a user by their username or email
+   * @param {string} username
```
gshaw1997: Can we update identifier here as well

src/shared/interfaces/DataStore.ts
```diff
@@ -171,6 +204,14 @@ export interface DataStore
...
+   * @returns {Promise<string>}
+   * @memberof DataStore
+   */
+  findUserId(username: string): Promise<string>;
```
gshaw1997: Can we update identifier here as well

PR #404, opened 2019-07-09
Patch artifact creation during Learning Object release process

gshaw1997:
**Purpose of this PR**
Addresses the problem describe in this story https://app.clubhouse.io/clarkcan/story/767/recently-released-learning-objects-not-appearing.

**Changes**
- Retrieve the unreleased file metadata from the `FileMetadata` module
- Create `CCLicenseRetriever` to fetch a `Readable` stream of the Creative Commons License within the interactor instead of the `ArchiverBundler`
- Create `FileGateway` to fetch a `Readable` stream of the Learning Object's files from the `FileManagerModule` within the interactor instead of the `ArchiverBundler`
- Handle stream error causing execution to terminate by adding event listener for `error` event and rejecting the error when fetching the Creative Commons License via the `https` library.
- Provide `CCLicenseRetriever` and `FileGateway` as dependencies to allow swapping for stubs during tests.

**WIP**
 This PR is dependent on https://github.com/Cyber4All/learning-object-service/pull/396

Once the above PR is merged the TODO in the `ModuleFileGateway` to make the call to the `streamFile` method of the `FileManagerModule` must be taken care of before this patch can be merged.

**Note on failing tests**
- The PATCH request tests to update a Learning Object are failing in the `LearningObjectsRouteHandler.spec.ts` file because this PR is using the `FileMetadata` module directly in the `LearningObjectInteractor` which isn't initialized in the test suite. 
This issue is resolved in PR https://github.com/Cyber4All/learning-object-service/pull/396 by using a gateway to interact with the module instead of calling the module directly which allows a stub to be used in testing. Once #396 is merged and integrated into this branch, the issue should be resolved.


src/LearningObjects/Publishing/Bundler/FileGateway.ts
```diff
@@ -0,0 +1,19 @@
+import { Readable } from 'stream';
+
+export abstract class FileGateway {
+  /**
+   * Retrieves a Readable stream if a user's Learning Object's file
```
gshaw1997:
```suggestion
   * Retrieves a Readable stream of a user's Learning Object's file
```

src/LearningObjects/Publishing/Bundler/CCLicenseRetriever.ts
```diff
@@ -0,0 +1,12 @@
+import { Readable } from 'stream';
+
+export abstract class CCLicenseRetriever {
+  /**
+   * Returns a readable stream of the latest creative commons license
```
gshaw1997:
```suggestion
   * Returns a readable stream of the latest Creative Commons license
```

gshaw1997:
> Everything looks good, though I am a little unclear on what the purpose of specifying that the CC license is the `latest` is. @gshaw1997 could you elaborate on that?

My thought process was, there are multiple versions of the Creative Commons License, so the function name should be declarative of the fact that the CC License that is returned would be the latest version used.

DonSeannelly:
> > Everything looks good, though I am a little unclear on what the purpose of specifying that the CC license is the `latest` is. @gshaw1997 could you elaborate on that?
> 
> @DonSeannelly My thought process was, there are multiple versions of the Creative Commons License, so the function name should be declarative of the fact that the CC License that is returned would be the latest version used for Learning Objects.

Ah, I understand now.  The only issue with this is that Creative Commons licenses are not exactly a linear progression. There’s a ton of different derivative types and we’re targeting a very specific one. I’m honestly thinking that we should remove mention of Creative Commons from the code, and just specify that it’s a license

PR #405, opened 2019-07-09
Allow users to filter by released and not released learning objects

davidnso:
This PR handles the following cases for searching and filtering a user's learning objects.

    If the requester is not the author and is not a privileged user, only summaries of released Learning Objects should be returned. The not released filter cannot be applied in this case.

    If the requester is the author, and has not specified a filter, summaries of released and all not released Learning Object should be returned.

    If they have specified filtering by released, only summaries of released Learning Objects should be returned.
    If they have specified filtering by not released, only summaries of not released Learning Objects should be returned.

    If the requester is a curator/reviewer of a collection/collections of the user's Learning Objects and has not specified a filter, summaries of released and review stage Learning Objects in their collection should be returned.
    If they have specified filtering by released, only summaries of released Learning Objects should be returned.
    If they have specified filtering by not released, only summaries of review stage in their collection Learning Objects should be returned.

    If the requester is a admin/editor and has not specified a filter, summaries of released and all review stage Learning Objects should be returned.
    If they have specified filtering by released, only summaries of released Learning Objects should be returned.
    If they have specified filtering by not released, only summaries of review stage Learning Objects should be returned.


gshaw1997:
This PR is inaccurate and does not contain the original changes from https://github.com/Cyber4All/learning-object-service/pull/399

PR #406, opened 2019-07-10
Allow users to have read access to released change logs

nvisal1:
***Purpose of this PR***
Allow users to have read access to change logs that belong to released Learning Objects.
This PR is to allow change logs to be available on the details page for logged in users.

This PR will cover the following cases:
1. if Learning Object is released and does not have a revision, fetch all change logs
2. if Learning Object has never been released (no revision), check user role, if privileged, fetch all change logs
3. if released Learning Object has a revision, but the user is looking at the released copy, fetch all change logs before last updated date on the released copy
4. if released Learning Object has revision, and the user is looking at the revision copy, check role, if privileged, fetch all change logs.

***Story***
This PR is in reference to Clubhouse story https://app.clubhouse.io/clarkcan/story/272/view-change-logs-from-details-page

***Implementation***
Case 1 and 2: can be determined by evaluating the `revision number` and `status` of a Learning Object
Case 3 and 4: can be determined by evaluating the `revision number` of a Learning Object and the `minusRevision` query parameter.

If the `minusRevision` query parameter is included, any change logs that were created after the last updated date on the Learning Object's released copy will not be included.

***Other Notable Changes***
The `fetchRecentChangelog` route has been deprecated. The `recent` query parameter has been added to the `fetchChangelogs` route. This was done in order to reduce confusion about the route structure. This is also more in-line with the idea of using query parameters to return a subset of the plain endpoint response.


src/Changelogs/ChangelogInteractor.ts
```diff
@@ -114,31 +117,75 @@ export async function getChangelogs(params: {
...
-  const isOwnedByAuthor = await params.dataStore.checkLearningObjectExistence({
+
+  return role;
+}
+
+/**
+ * - Determines if the request to read change logs is valid
+ * This request validation is performed by checking that the given
+ * authorId and learningObjectId pair exist.
+ * - Ensure that the user making the request has access to modify changelogs
```
DonSeannelly: It looks like `modify` may have been a copy and paste error?

src/Changelogs/ChangelogInteractor.ts
```diff
@@ -114,31 +117,75 @@ export async function getChangelogs(params: {
...
+ * If the Learning Object is released, all users have access to read change logs.
```
DonSeannelly: Does this mean that visitors to the site will not be able to read change logs?

nvisal1: Yes, the story states "As a user". I thought that meant that users needed to be logged in.

DonSeannelly: Which story is this in reference to?

nvisal1: https://app.clubhouse.io/clarkcan/story/272/view-change-logs-from-details-page

src/Changelogs/interfaces/ChangelogDataStore.ts
```diff
@@ -0,0 +1,36 @@
+import { ChangeLogDocument } from '../../shared/types/changelog';
+
+export interface ChangelogDataStore {
+    createChangelog(params: {
```
gshaw1997: Document interface

src/Changelogs/testing/StubModuleLearningObjectGateway.ts
```diff
@@ -0,0 +1,20 @@
+import { LearningObjectGateway } from '../LearningObjectGateway';
+import { Requester } from '../typings';
+import { LearningObjectSummary } from '../../shared/types';
+import { Stubs } from '../../tests/stubs';
+import { LearningObject } from '../../shared/entity';
+import { STUB_CHANGELOG_IDS } from './ChangelogStubs';
+
+export class StubModuleLearningObjectGateway extends LearningObjectGateway {
```
gshaw1997: Use implements not extends. `StubModuleLearningObjectGateway` isnt a sub-class
```suggestion
export class StubModuleLearningObjectGateway implements LearningObjectGateway {
```

gshaw1997: Also, couldn't we just rename to `StubLearningObjectGateway` why is it necessary to specify `Module`?

src/Changelogs/ModuleLearningObjectGateway.ts
```diff
@@ -0,0 +1,22 @@
+import { LearningObjectGateway } from './LearningObjectGateway';
+import { LearningObjectAdapter } from '../LearningObjects/adapters/LearningObjectAdapter';
+import { LearningObjectSummary } from '../shared/types';
+import { Requester } from './typings';
+
+export class ModuleLearningObjectGateway extends LearningObjectGateway {
```
gshaw1997: Use `implements` not `extends`. `ModuleLearningObjectGateway` is not a sub-class.
```suggestion
export class ModuleLearningObjectGateway implements LearningObjectGateway {
```

src/Changelogs/ChangelogInteractor.ts
```diff
@@ -59,53 +62,169 @@ export async function getRecentChangelog(params: {
...
+ * the return type of this function will change to ChangeLogDocument instead of
+ * ChangeLogDocument[]
+ * @param {optional parameter - boolean} minusRevision if true, the function will
+ * return all change logs that are relevant to the released copy of the specified Learning
+ * Object. This excludes all change logs that were created after the date property on the
+ * Release copy.
+ * @returns {ChangeLogDocument[] | ChangeLogDocument} This changes beased on the recent parameter
  */
 export async function getChangelogs(params: {
+  learningObjectGateway: LearningObjectGateway;
```
jbarlow96: Add as a param in JSDoc

src/Changelogs/testing/StubModuleLearningObjectGateway.ts
nvisal1: I have seen other modules follow the naming convention of interface = <name>gateway and concretion = module<name>gateway. I was just trying to be consistent. Thoughts?

PR #408, opened 2019-07-11
Fix Learning Object children not being fully loaded when requesting parent

gshaw1997:
**Purpose of this PR**
When requesting a full Learning Object, its children were not fully loaded. This PR fixes that issue by loading all of the parent's children fully.

**Changes**
- Load full child objects recursively when requesting a parent Learning Object.


gshaw1997:
After a discussion with the @Cyber4All/backend team, we have decided to change the `LearningObject` resource that is returned from the `/users/:username/learning-objects/:learningObjectId`, `/users/:username/learning-objects/:learningObjectName`, and `/learning-objects/:username/:learningObjectName` routes to only return `LearningObjectChildSummary`s in the `children` property instead of full child Learning Objects. In order to fetch the full child Learning Object, the child Learning Object must be explicitly requested using one of the above routes. This decision was made to conserve service resources when requesting Learning Objects with a deeply nested child structure.

PR #413, opened 2019-07-16
Add route for creating Learning Object revisions

nvisal1:
***Purpose***
The purpose of this PR is to add a route for creating Learning Object revisions.

***Story***
This PR is in response to story https://app.clubhouse.io/clarkcan/story/787/create-a-route-for-making-new-revisions

***Implementation***
The new route is structured as
`POST /users/:userId/learning-objects/:id/revisions`

Once the above route is hit, the following steps occur: 
1. Validate request by checking that the given `userId` and `learningObjectId` produce a Learning Object
2. Retrieve the released copy of the Learning Object
3. Check that the requester is the Learning Object Author
4. Update the working copy Learning Object by setting status to `unreleased` and
revision to `workingCopy.revision++`.


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1330,6 +1330,74 @@ export async function getMaterials({
...
+
+  const releasedCopy = await getReleasedLearningObjectSummary({
+    dataStore: params.dataStore,
+    id: params.learningObjectId,
+  });
+
+  if (!releasedCopy) {
+    throw new ResourceError(
+      `Learning Object with id ${params.learningObjectId} is not released`,
+      ResourceErrorReason.CONFLICT,
```
gshaw1997: Why `CONFLICT` here shouldnt it be `BAD_REQUEST`

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1330,6 +1330,74 @@ export async function getMaterials({
...
+  });
+
+  const releasedCopy = await getReleasedLearningObjectSummary({
+    dataStore: params.dataStore,
+    id: params.learningObjectId,
+  });
+
+  if (!releasedCopy) {
+    throw new ResourceError(
+      `Learning Object with id ${params.learningObjectId} is not released`,
```
gshaw1997:
```suggestion
      `Cannot create a revision of Learning Object: ${params.learningObjectId} since it is not released.`,
```

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1330,6 +1330,74 @@ export async function getMaterials({
...
+      ResourceErrorReason.CONFLICT,
+    );
+  }
+
+  if (releasedCopy.author.username !== params.requester.username) {
+    throw new ResourceError(
+      `Requester ${params.requester.username} does not own Learning Object with id ${params.learningObjectId}`,
```
gshaw1997:
```suggestion
      `Cannot create a revision. Requester ${params.requester.username} must be the author of Learning Object with id ${params.learningObjectId}`,
```

src/LearningObjects/LearningObjectInteractor.ts
nvisal1: It should be `bad requests. Good catch.

nvisal1:
Update: I needed to change the logic in the AuthorziationManager in order to update released Learning Objects

src/shared/AuthorizationManager.ts
```diff
@@ -394,14 +396,20 @@ export function authorizeWriteAccess({
     authorUsername: learningObject.author.username,
     requester,
   });
-  const authorAccess = isAuthor && isUnreleased;
+  let isRevision = false;
+  if (updates && updates.revision) {
```
jbarlow96: Confused by this If statement. Why does it have both updates && updates revision? Not saying it's wrong, just want to make sure that I'm on the same page.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1453,6 +1454,76 @@ export async function getMaterials({
...
+  if (releasedCopy.author.username !== params.requester.username) {
+    throw new ResourceError(
+      `Cannot create a revision. Requester ${params.requester.username} must be the author of Learning Object with id ${params.learningObjectId}`,
+      ResourceErrorReason.INVALID_ACCESS,
+    );
+  }
+
+  releasedCopy.revision++;
+
+  await updateLearningObject({
```
gshaw1997: We're baking a lot of functionallity into the updateLearningObject function which is causing some heavy coupling and redundancy in fetching data since it has its own validation logic. I think we should think of creating a revision as its own process and have this function be responsible for authorization and update the required properties directly using the datastore.

src/shared/AuthorizationManager.ts
```diff
@@ -381,10 +381,12 @@ export function authorizeWriteAccess({
   learningObject,
   requester,
   message,
+  updates,
```
gshaw1997: I dont think it is a good idea for us to add to the complexity of this function. I think it should be left to authorizing write access to Learning Object metadata only soley based on the user's permissions over the Learning Object while its in different states, not based on the update operation.

src/shared/types/learning-object-updates.ts
```diff
@@ -11,6 +11,7 @@ export interface LearningObjectUpdates {
   published?: boolean;
   collection?: string;
   materials?: { [index: string]: any };
+  revision?: number;
```
gshaw1997: I think we should reserve the `LearningObjectUpdate` type for Learning Object metadata that can be updated by users directly. `revision` is really just a property used for the system to understand certain information about a Learning Object. A users should never be able to update this directly.

src/shared/types/learning-object-updates.ts
```diff
@@ -26,6 +27,7 @@ const LearningObjectUpdateProps: LearningObjectUpdates = {
   status: '',
   published: false,
   collection: '',
+  revision: 0,
```
gshaw1997: I think we should reserve the LearningObjectUpdate type for Learning Object metadata that can be updated by users directly. revision is really just a property used for the system to understand certain information about a Learning Object. A users should never be able to update this directly.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -950,6 +950,7 @@ export async function updateLearningObject({
     authorizeWriteAccess({
       learningObject,
       requester,
+      updates,
```
gshaw1997: I dont think it is a good idea for us to add to the complexity of this function. I think it should be left to authorizing write access to the Learning Object metadata that is updatable by a user soley based on the user's permissions over the Learning Object while its in different states, not based on the properties being updated.

src/shared/AuthorizationManager.ts
nvisal1: The conditional checks to ensure that `updates` is truthy (since it is an optional parameter) and then checks to make sure that the requester is updating the `revision` property.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1453,6 +1453,75 @@ export async function getMaterials({
...
+  ) {
+    throw new ResourceError(
+      `Cannot create a revision. Requester ${params.requester.username} must be the author of Learning Object with id ${params.learningObjectId}`,
+      ResourceErrorReason.INVALID_ACCESS,
+    );
+  }
+
+  releasedCopy.revision++;
+
+  await params.dataStore.createRevision({
```
gshaw1997: What was the reasoning for creating a method in the datastore that does this? 

To me, creating a revision is a business logic level detail, so when an action is triggered to create a revision the biz-logic should call out to the necessary driver and gateway methods to do what we understand as creating a revision, which is just updating the revision id and status. Putting this action at the datastore level means that all datastores need to understand what a revision is, instead of just understanding how to fetch, update, and delete data.

nvisal1: I created a new datastore method because the editLearningObject function accepts type `LearningObjectUpdates`. It was stated earlier that we did not want the `revision` property to be included in the `LearningObjectUpdates` type. Thoughts? @gshaw1997

gshaw1997: Can we split out the types and have the updates type the datastore method takes be a superset of the user editable metadata?

src/drivers/MongoDriver.ts
```diff
@@ -1,4 +1,4 @@
-import { Cursor, Db, MongoClient, ObjectID } from 'mongodb';
+import { Cursor, Db, MongoClient, ObjectID, UpdateWriteOpResult } from 'mongodb';
```
gshaw1997: What's this import?

PR #429, opened 2019-07-25
Separate Revisions from Core Learning Object Logic

DonSeannelly:
**This PR is not yet complete, but with making a few big refactorings to the existing Revision resource I wanted to get some eyes on before I add code to complete this story.**

This commit takes the ability to read and create revisions and moves it
into a subdirectory of LearningObjects, with its own interactor and
RouteHandler. This gives the RESTful Resource a more distinctive home in
the project and better adheres to SRP (changes to Revision logic won't
require us to muck around with other Learning Object functionality).

Additionally, tests for the LearningObjectAdapter that were testing
Revision logic have been converted to interactor-level tests in the
migration, since the purpose of testing the adapter was in fact to
validate that the interactor worked as expected.


DonSeannelly:
EDIT:

The PR is feature complete, and takes care of a bug allowing multiple revisions to be made by an Author.

PR #430, opened 2019-07-29
Update Learning objects in Elasticsearch and Mongo simultaneously. 

davidnso:
The purpose of this PR is to allow admins and editors to make changes to submitted learning objects in the elastic search index while those changes are also made in mongo. This is done by grabbing an instance of the ```learningObjectAdapter``` in the ```LearningObjectInteractor``` and calling the update function.
This PR also improves validation logic and Error handling when a learning object name is being changed.


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1635,9 +1646,14 @@ function validateUpdates(params: {
   id: string;
   updates: LearningObjectMetadataUpdates;
 }): void {
-  if (params.updates.name) {
-    if (params.updates.name.trim() === '') {
-      throw new Error('Learning Object name cannot be empty.');
+ if (params.updates.name) {
+  const learningObjectName = params.updates.name.trim();
+  const isValidName = learningObjectName.length >= 2 && learningObjectName.length < 50;
```
DonSeannelly: Does the Learning Object entity type have this kind of validation baked in?

davidnso: Yes it does, but that validation is private to the entity so I don't think I'd be able to use that here. And entity also checks the max length for anything less than 170. But in the builder the message says a name must be less than 50 characters. Which one is valid?

gshaw1997: @davidnso The Learning Object entity class is the source of truth for validation, there should be a method exposed for validating name

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
```diff
@@ -46,6 +46,47 @@ export class ElasticsearchSubmissionPublisher implements SubmissionPublisher {
...
+  async updateSubmission(params: {
+    learningObjectId: string;
+    updates: Partial<LearningObject>;
+  }) {
+    const { learningObjectId, updates } = params;
+    let updateField = Object.keys(updates);
+    let updateSource: string = '';
+    updateField.map(field => {
+       let updateValue = updates[field];
+       updateSource = updateSource.concat(`ctx._source.${field} = \"${updateValue}\";`);
```
gshaw1997: Why `concat` here instead of just using `+=`?

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -795,15 +800,19 @@ export async function updateLearningObject({
       }.`,
     });
     const cleanUpdates = sanitizeUpdates(updates);
-    validateUpdates({
```
gshaw1997: Why was this removed?

davidnso: I added the validation check inside `sanitizeUpdates` instead since it didn't make sense to sanitize then validate in separate steps.

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1491,6 +1501,7 @@ function sanitizeUpdates(
       updates[key] = typeof value === 'string' ? value.trim() : value;
     }
   }
+  validateUpdates(updates);
```
gshaw1997: We should keep these functions separate, they have two different purposes.

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
davidnso: No real reason, It seemed cleaner to use concat

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
```diff
@@ -46,6 +46,47 @@ export class ElasticsearchSubmissionPublisher implements SubmissionPublisher {
...
+   * in storage.
+   * @inheritdoc
+   */
+  async updateSubmission(params: {
+    learningObjectId: string;
+    updates: Partial<LearningObject>;
+  }) {
+    const { learningObjectId, updates } = params;
+    let updateField = Object.keys(updates);
+    let updateSource: string = '';
```
DonSeannelly: Since a default value is assigned here, TypeScript favors the convention of omitting the type annotation:

```suggestion
    let updateSource = '';
```

I don't think our current linter complains, but I figured I'd still mention it.

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
```diff
@@ -46,6 +46,47 @@ export class ElasticsearchSubmissionPublisher implements SubmissionPublisher {
...
+      },
+    };
+    try {
+    await this.client.updateByQuery({
+        index: INDEX_NAME,
+        type: '_doc',
+        body: learningObjectUpdateRequest,
+      });
+    } catch (e) {
+      console.log(e);
```
DonSeannelly: What directed your use of a console log here instead of `reportError`?

src/LearningObjectSubmission/interactors/index.ts
```diff
@@ -1,5 +1,6 @@
 import { cancelSubmission } from './cancelSubmission';
 import { checkFirstSubmission } from './checkFirstSubmission';
 import { submitForReview } from './submitForReview';
+import {updateSubmission} from './updateSubmission';
```
DonSeannelly: Little nitpick here for consistency:
```suggestion
import { updateSubmission } from './updateSubmission';
```

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
davidnso: I'll change this in the next commit

src/shared/entity/learning-object/validators/learning-object.validator.ts
```diff
@@ -0,0 +1,399 @@
...
+import { ResourceError, ResourceErrorReason } from '../../../errors';
+
+/**
+ * Validates object is a valid Learning Object
+ *
+ * @export
+ * @param {LearningObject} object
+ * @returns {(void | never)}
+ */
+export function validate(object: LearningObject): void | never {
```
nvisal1: To improve readability, we should rename this function. Something like `validateLearningObject`.

src/shared/entity/learning-object/validators/learning-object.validator.ts
```diff
@@ -0,0 +1,399 @@
...
+  validateLength(object.length);
+  validateLevels(object.levels);
+  validateMaterials(object.materials);
+  validateMetrics(object.metrics);
+  validateName(object.name);
+  validateOutcomes(object.outcomes);
+  validateStatus(object.status);
+}
+
+export function validateUpdates(object: LearningObjectMetadataUpdates) {
```
nvisal1: `validateLearningObjectUpdates`?

src/shared/entity/learning-object/validators/learning-object.validator.ts
```diff
@@ -0,0 +1,399 @@
...
+  }
+}
+/**
+ * Validates object is a submittable Learning Object
+ *
+ * @export
+ * @param {LearningObject} object
+ * @returns {(void | never)}
+ */
+export function validateSubmittable(object: LearningObject): void | never {
```
nvisal1: `validateLearningObjectSubmission`?

src/shared/entity/learning-outcome/validators/learning-outcome.validator.ts
```diff
@@ -0,0 +1,100 @@
...
+import { StandardOutcome } from '../../standard-outcome/standard-outcome';
+
+/**
+ * Validates object is LearningOutcome
+ *
+ * @export
+ * @param {LearningOutcome} outcome
+ * @returns {(void | never)}
+ */
+export function validate(outcome: LearningOutcome): void | never {
```
nvisal1: `validateLearningOutcome`?

src/shared/entity/learning-outcome/validators/learning-outcome.validator.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Validates object is a submittable LearningOutcome
+ *
+ * @export
+ * @param {LearningOutcome} outcome
+ * @returns {(void | never)}
+ */
+export function validateSubmittable(outcome: LearningOutcome): void | never {
```
nvisal1: `validateLearningOutcomeSubmission`?

src/LearningObjects/Publishing/ModuleLearningObjectSubmissionGateway.ts
```diff
@@ -1,18 +1,42 @@
...
+import { LearningObject } from '../../shared/entity';
 
-export class ModuleLearningObjectSubmissionGateway implements LearningObjectSubmissionGateway {
+export class ModuleLearningObjectSubmissionGateway
+  implements LearningObjectSubmissionGateway {
+  /**
+   *
+   * @inheritdoc
+   */
+  deleteSubmission(params: {
```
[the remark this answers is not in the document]
DonSeannelly: @nvisal1 are you suggesting that we rename this function to match that of the function that it adapts? Personally, I would argue that expanding the name is redundant, since it is a function of the `ModuleLearningObjectSubmissionGateway `, the subject of the action is clear when we call `ModuleLearningObjectSubmissionGateway.deleteSubmission(...)`.

By the same argument, if we were to make changes, I would suggest changing the `LearningObjectSubmissionAdapter` function's name instead of this one.

src/LearningObjectSubmission/interactors/updateSubmission.ts
```diff
@@ -0,0 +1,30 @@
+import { SubmissionPublisher } from './SubmissionPublisher';
+import { UserToken, LearningObjectMetadataUpdates } from '../../shared/types';
+import { requesterIsAdminOrEditor } from '../../shared/AuthorizationManager';
+import { ResourceError, ResourceErrorReason } from '../../shared/errors';
+
+/**
+ * updateSubmission allows admins and editors to publish changes to a submission for other
+ * reviewers to see, as a Learning Object is moving through the submission pipeline.
+ */
+export async function updateSubmission(params: {
```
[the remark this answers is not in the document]
DonSeannelly: @nvisal1 can you elaborate on why the name of this should change? I'm not sure if I'm convinced yet.

As I see it, it's the home of business logic for the task of updating a submission. It performs authorization and then calls a driver to do the work via a shared interface. This seems to be the same pattern we use for all interactors.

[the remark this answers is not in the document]
DonSeannelly: > Can we use the ServiceModule package to resolve the `SubmissionPublisher` within the interactor in order to reduce the number of function arguments?

I think we might consider that in the future, but I'm hesitant to change everything over to one particular library because we seem to be changing our approaches every few weeks. My vote would be to hold off and maybe convert it in the future if we still like the library and we have additional work to be done in this module.

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
```diff
@@ -46,6 +47,52 @@ export class ElasticsearchSubmissionPublisher implements SubmissionPublisher {
...
   }
 
+  /**
+   * @description
+   * In the case that an update does fail, an error will be reported for us to fix retrospectively.
+   * Until that is done, the searchable submissions will be out of date with the Learning Object data
+   * in storage.
+   * @inheritdoc
+   */
+  async updateSubmission(params: {
```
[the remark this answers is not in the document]
DonSeannelly: I agree with @nvisal1, the code has quite a bit of nesting and complexity. Some helper methods would make it much more approachable.

src/LearningObjectSubmission/ElasticsearchSubmissionPublisher.ts
```diff
@@ -55,14 +56,19 @@ export class ElasticsearchSubmissionPublisher implements SubmissionPublisher {
    */
   async updateSubmission(params: {
     learningObjectId: string;
-    updates: Partial<LearningObject>;
+    updates: LearningObjectMetadataUpdates;
```
DonSeannelly: Thank you for introducing this type! It makes the code easier to reason about 👍

src/shared/entity/learning-object/validators/learning-object.validator.ts
davidnso: Yea I like that

src/shared/entity/learning-object/validators/learning-object.validator.ts
davidnso: I'd say `validateSubmittableLearningObject`

src/shared/entity/learning-outcome/validators/learning-outcome.validator.ts
davidnso: how about `validateSubmittableLearningOutcome`?

src/LearningObjectSubmission/adapters/LearningObjectSubmissionAdapter.ts
```diff
@@ -1,47 +1,66 @@
...
+      learningObjectId: params.learningObjectId,
+      authorUsername: params.authorUsername,
+      user: params.user,
+    });
+  }
+
+  /**
+   * Calls the updateSubmission interactor function.
+   */
+  updateLearningObjectSubmission({
```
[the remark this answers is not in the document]
davidnso: The name was kept this way to match the pattern used in the `deleteLearningObjectSubmission` function. I'm more on the side of keeping things consistent here especially since `updateSubmission` is a function used in quite a few files. Just so that our entry point in the adapter is easily distinguishable from the interactor function and all other calls there after.

src/LearningObjectSubmission/interactors/updateSubmission.ts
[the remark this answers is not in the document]
nvisal1: I agree with the point about the Service Module package, but I would still like to see the name change. It is confusing to have two functions with the same name to perform different tasks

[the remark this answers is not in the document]
DonSeannelly: > I agree with the point about the Service Module package, but I would still like to see the name change. It is confusing to have two functions with the same name to perform different tasks

 What would you suggest as far as the name change goes?

src/LearningObjects/Publishing/ModuleLearningObjectSubmissionGateway.ts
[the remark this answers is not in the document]
nvisal1: I like the idea of having gateways, adapters and interactor functions map to the same name. I would suggest changing all three to `deleteSubmission`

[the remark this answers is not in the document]
DonSeannelly: I support @nvisal1’s suggestion for the change

src/LearningObjectSubmission/adapters/LearningObjectSubmissionAdapter.ts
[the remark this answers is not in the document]
nvisal1: I like the idea of having gateway, adapter and interactor functions with shared names. I do not think that making those functions easily distinguishable buys us anything since outside modules can only access the adapter function.

src/LearningObjectSubmission/interactors/updateSubmission.ts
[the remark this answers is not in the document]
nvisal1: I would suggest keeping the business logic function name the same and changing the publisher function to `applySubmissionUpdates`. The updates are not actually applied until validation occurs. Thoughts? Naming is hard...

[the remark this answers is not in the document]
DonSeannelly: Yeah, naming is nothing but a series of million dollar questions. I like the notion of specifying that we are applying or saving updates. I could also see calling it something like publishSubmissionUpdates. @davidnso I think you should make a judgement call on the name, but I think we’re set on it changing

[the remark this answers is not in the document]
davidnso: I'm leaning towards using `applySubmissionUpdates`

PR #436, opened 2019-08-02
Optimize the release and download process 

davidnso:
This PR... 

-  Introduces the `HierarchicalLearningObject` type which contains the full hierarchy of a Learning Object; the parent and children at every level are full Learning Objects.
- Changes the release process to generate a `HierarchicalLearningObject` which is then passed to the publishing service. 
- Removes unnecessary properties from the `LearningObjectChildSummary` type.


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -938,15 +941,15 @@ export async function updateLearningObject({
  * @param {UserToken} requester [The requester of the releasable Learning Object]
  * @returns {Promise<LearningObject>}
  */
-async function generateReleasableLearningObject({
+export async function generateReleasableLearningObject({
```
gshaw1997: Do we need this function exported?
```suggestion
async function generateReleasableLearningObject({
```

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1029,22 +1032,18 @@ export async function getLearningObjectById({
...
+
+    let children: LearningObjectSummary[];
     if (loadingReleased) {
       learningObject.materials.files = learningObject.materials.files.map(
         appendFilePreviewUrls(learningObject),
       );
-      children = await dataStore.loadReleasedChildObjects({
-        id: learningObject.id,
-        full: false,
+      children = await loadChildObjectSummaries({
```
gshaw1997: Why did we remove the call to the datastore here? Does the `loadChildSummaries` method load from the `released-objects` collection or does it query the `objects`collection for released Learning Objects. We need to be sure we're actually querying the `released-objects` collection

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1164,21 +1164,11 @@ async function loadChildObjectSummaries({
...
-      child.materials.files = await Gateways.fileMetadata().getAllFileMetadata({
-        requester,
-        learningObjectId: child.id,
-        filter: 'unreleased',
-      });
-      return child;
-    }),
-  );
-  return children;
+  let childSummary = children.map(child => {
```
gshaw1997: We can simplify this to
```suggestion
  let childSummary = children.map(mapLearningObjectToSummary)
```

src/LearningObjects/Publishing/Bundler/Interactor.ts
```diff
@@ -163,19 +164,19 @@ function addFiles({
...
   children,
   prefix = '',
 }: {
-  children: LearningObject[];
+  children: HierarchicalLearningObject[];
   prefix: string;
 }): Promise<BundleData[]> {
   const childrenBundleData = await Promise.all(
-    children.map(child => {
+    children.map(async child => {
```
gshaw1997: Async not needed here
```suggestion
    children.map(child => {
```

src/LearningObjects/LearningObjectInteractor.ts
davidnso: it's loading from the objects collection, This will be changed in the next commit.

src/LearningObjects/LearningObjectInteractor.ts
davidnso: In order to test whether or not this object is returning the full hierarchy of a Learning Object I thought it'd be necessary to export this function to make it available in the spec file.

src/LearningObjects/Publishing/Bundler/Interactor.ts
davidnso: The `loadReleasedChildObjects` function returns a Promise so removing the async label seems to break it.

PR #440, opened 2019-08-05
Return new revision id when revision is created

nvisal1:
***Purpose***
The purpose of this PR is to provide the client with the new Learning Object revision id after a revision is created.

***Implementation***
The route:
`POST /users/:username/learning-objects/:learningObjectId/revisions`
will now return a response of `200 {newRevisionId: number}` when successful.

***Story***
This PR is in response to Clubhouse story https://app.clubhouse.io/clarkcan/story/945/return-new-revision-id-after-creating-a-new-revision-in-learning-object-service


src/LearningObjects/Revisions/RevisionRouteHandler.ts
```diff
@@ -26,8 +26,8 @@ export function initializePrivate({
   const createRevision = async (req: Request, res: Response) => {
     try {
       const params = { ...req.params, dataStore, requester: req.user };
-      await RevisionInteractor.createLearningObjectRevision(params);
-      res.sendStatus(204);
+      const newRevisionId = await RevisionInteractor.createLearningObjectRevision(params);
+      res.status(200).json({newRevisionId});
```
gshaw1997: Why not just `revision` since this maps up to a property that exists on the entity?
```suggestion
      res.status(200).json({revision: newRevisionId});
```

PR #446, opened 2019-08-07
Fix: Publishing Revisions Creates a Duplicate in the Index

nvisal1:
***Purpose***
This PR fixes the issue where publishing revisions would create a duplicate in the Elasticsearch index.

***Implementation***
* When a Learning Object is released, Elasticsearch will now attempt to update the existing Released Learning Object document. If the document does not exist, the releasable object is indexed.
* A new Elasticsearch helper function was created: `formatUpdateQueryParam` - This function takes an object and returns a formatted string that is accepted by the Elasticsearch updateByQuery API.

***Story***
This PR responds to Clubhouse story https://app.clubhouse.io/clarkcan/story/896/publishing-revisions-creates-a-duplicate-in-the-index


src/shared/elasticsearch/HelperFunctions.ts
```diff
@@ -27,3 +27,42 @@ export function cleanLearningObject(learningObject: LearningObject) {
...
+
+/**
+ * formatUpdateQueryParam takes an object and formats
+ * it into a string that is accepted by the Elasticsearch
+ * updateByQuery function
+ *
+ * @param cleanObject
+ */
+export function formatUpdateQueryParam(
+    cleanObject: {},
```
davidnso: We should give this object a specific type, I think it would be nice to create a new interface to match the documents that are being stored in elastic. I.e an ElasticLearningObjectDocument. @DonSeannelly What are your thoughts on this?

DonSeannelly: @davidnso I like the idea. Perhaps `SearchableLearningObject` or `LearningObjectSearchDocument`? I'm not solid on a name, but basically conveying that the data contained inside is indexable and what we want to use for Learning Object search.

In theory, this data type wouldn't change even if we moved to Algolia over Elasticsearch

davidnso: Yea, I like the sound of that. I'm leaning towards `LearningObjectSearchDocument`.

src/shared/elasticsearch/HelperFunctions.spec.ts
```diff
@@ -11,44 +12,114 @@ const stubs = new Stubs();
...
+            level: 2,
+          },
+          {
+            name: 'test 2',
+            level: 3,
+          },
+        ],
+      };
+      // tslint:disable-next-line
+      const formattedString = "ctx._source.name = \"test\";ctx._source.level = \"2\";ctx._source.object[0].name = \"test\";ctx._source.object[0].level = \"2\";ctx._source.object[1].name = \"test 2\";ctx._source.object[1].level = \"3\";"
```
jbarlow96: Similar to what @DonSeannelly said about another part in this PR, would the use of back-ticks remove the tslint ignore comment here too? (as well as the ignores on lines 110 and 120)

src/LearningObjects/Publishing/ElasticSearchPublishingGateway.ts
```diff
@@ -23,11 +25,49 @@ export class ElasticSearchPublishingGateway implements PublishingDataStore {
   constructor() {
     this.client = new Client({ node: URI });
   }
+
+  /**
+   * addToReleased attempts to update the existing Released Learning
+   * Object document in Elasticsearch. If the Release Learning Object
```
DonSeannelly:
```suggestion
   * Object document in Elasticsearch. If the Released Learning Object
```

PR #448, opened 2019-08-08
Point search traffic to Elastic

davidnso:
This PR directs all user traffic to the Elastic search index, expect in the case where an author is searching/viewing objects in their dashboard.

### DIFF
--- a/src/interactors/LearningObjectInteractor.ts
+++ b/src/interactors/LearningObjectInteractor.ts
@@ -420,7 +420,7 @@ export class LearningObjectInteractor {
         status,
       } = this.formatSearchQuery(query);
       let response: { total: number; objects: LearningObject[] };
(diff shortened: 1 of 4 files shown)


src/interactors/LearningObjectInteractor.ts
```diff
@@ -420,7 +420,7 @@ export class LearningObjectInteractor {
         status,
       } = this.formatSearchQuery(query);
       let response: { total: number; objects: LearningObject[] };
-
+      console.log('Hitting the learning object interactor');
```
nvisal1: Remove dev log

davidnso: Good catch!

nvisal1:
> Is the PR that @jbarlow96 is working on going to remove what `next()` was previously passing the request on to?

Yes it will

PR #450, opened 2019-08-13
Fix isFirstSubmission Bug

nvisal1:
***Purpose***
This PR fixes the route that checks if a Learning Object has been submitted to a collection before. This fixes the issue where changelogs are enforced on first-time submissions.

***Implementation***
The `hasSubmission` function in the `MongoSubmissionDatastore` was querying for the wrong document property. Because of that, the result was always undefined. The `checkFirstSubmission` interactor function now returns the inverse of the result from `hasSubmission`.

***Story***
This PR responds to Clubhouse story https://app.clubhouse.io/clarkcan/story/1005/fix-isfirstsubmission-bug


src/LearningObjectSubmission/MongoSubmissionDatastore.ts
```diff
@@ -92,12 +92,20 @@ export class MongoSubmissionDatastore implements SubmissionDataStore {
...
+        {
+          learningObjectId: params.learningObjectId,
+          collection: params.collection,
+        },
+        { projection: { _id: 1 }})
       .limit(1)
-      .toArray()[0];
-    return submission !== undefined;
+      .toArray();
+    return submission[0] !== undefined;
```
DonSeannelly: Since this property is of type Array, the name should be pluralized to reflect that

PR #456, opened 2019-08-21
Generate File paths with file access ids

nvisal1:
This PR fixes issues that existed in https://github.com/Cyber4All/learning-object-service/pull/452, which was reverted.

Fixed issues:
* Set new module routes in ExpressRouteDriver
* Return error object from file access id mongo driver when no documents are found
* Elasticsearch test runner container timeout


src/FileAccessIdentities/adapters/FileAccessIdentitiesAdapter/FileAccessIdentitiesAdapter.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  static open() {
+    FileAccessIdentitiesAdapter._instance = new FileAccessIdentitiesAdapter();
+  }
+
+  static getInstance(): FileAccessIdentitiesAdapter {
+    if (this._instance) {
+      return this._instance;
+    }
+    throw new Error(
```
nrwinner: This is an interesting pattern here. Why not, in the event that an instance hasn't been created, simply create it and return the instance?

nvisal1: We throw an error here because all adapters are initialized in app.ts at start time. If this error occurs, it it is because the developer failed to follow the pattern correctly. This is for maintaining code standards.

nrwinner: Ahh good stuff.

PR #459, opened 2019-08-23
Implement new route to fetch the Learning Outcomes from a Learning Object

nrwinner:
This PR introduces a new public route for fetching the Learning Outcomes from a specified Learning Object. **It's a WIP because it still needs unit tests, but I would like a review of the route before I implement said tests.**

### DIFF
--- a/src/LearningOutcomes/LearningOutcomeInteractor.ts
+++ b/src/LearningOutcomes/LearningOutcomeInteractor.ts
@@ -1,5 +1,4 @@
 import {
-  LearningOutcomeInput,
   LearningOutcomeInsert,
   LearningOutcomeUpdate,
 } from './types';
@@ -7,6 +6,7 @@ import { UserToken } from '../shared/types';
 import { sanitizeObject } from '../shared/functions';
 import { LearningOutcome } from '../shared/entity';
 import { ResourceError, ResourceErrorReason } from '../shared/errors';
+import { LearningObjectGateway } from './gateways/ModuleLearningObjectGateway';
 
 export interface LearningOutcomeDatastore {
   insertLearningOutcome(params: {
@@ -42,7 +42,7 @@ export async function addLearningOutcome(params: {
   source: string;
   outcomeInput: Partial<LearningOutcome>;
 }): Promise<string> {
-  const {dataStore, user, source, outcomeInput} = params;
+  const { dataStore, user, source, outcomeInput } = params;
   let outcome: Partial<LearningOutcome>;
   try {
     outcome = new LearningOutcome(outcomeInput);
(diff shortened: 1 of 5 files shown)


src/LearningOutcomes/LearningOutcomeInteractor.ts
```diff
@@ -80,6 +80,39 @@ export async function getLearningOutcome(params: {
...
+  dataStore,
+  requester,
+  learningObjectGateway,
+  source,
+}: {
+  dataStore: LearningOutcomeDatastore,
+  requester: UserToken,
+  learningObjectGateway: LearningObjectGateway,
+  source: string,
+}) {
```
nvisal1: Missing return type

src/LearningOutcomes/LearningOutcomeInteractor.ts
```diff
@@ -80,6 +80,39 @@ export async function getLearningOutcome(params: {
...
+    // authorization/not-found errors are thrown in the LearningObjectAdapter and handled in the LearningOutcomeRouteHandler
```
nvisal1: Can you move this comment to the above JSDoc

src/LearningOutcomes/LearningOutcomeInteractor.ts
```diff
@@ -80,6 +80,39 @@ export async function getLearningOutcome(params: {
...
+    await learningObjectGateway.getLearningObject(source, requester);
```
nvisal1: I would prefer to handle the authorization/errors within this function. I think that it would significantly improve readability. Thoughts @DonSeannelly ?

src/LearningOutcomes/gateways/ModuleLearningObjectGateway.ts
```diff
@@ -0,0 +1,43 @@
...
+  async getLearningObject(learningObjectId: string, requester: UserToken): Promise<LearningObjectSummary> {
+    try {
+      let learningObject: LearningObjectSummary = await this.adapter.getReleasedLearningObjectSummary(learningObjectId);
+      return learningObject;
+    } catch (error) {
+      if (error instanceof ResourceError && error.name === ResourceErrorReason.NOT_FOUND) {
+        // handle the not found error and check for a working-copy
+        return await this.adapter.getWorkingLearningObjectSummary({ id: learningObjectId, requester });
+      } else {
+        // throw the error
```
nvisal1: I don't think we need this comment. `throw error` can be understood without it.

src/LearningOutcomes/gateways/ModuleLearningObjectGateway.ts
```diff
@@ -0,0 +1,43 @@
...
+   * @returns {Promise<LearningObjectSummary>}
+   * @memberof ModuleLearningObjectGateway
+   */
+  async getLearningObject(learningObjectId: string, requester: UserToken): Promise<LearningObjectSummary> {
+    try {
+      let learningObject: LearningObjectSummary = await this.adapter.getReleasedLearningObjectSummary(learningObjectId);
+      return learningObject;
+    } catch (error) {
+      if (error instanceof ResourceError && error.name === ResourceErrorReason.NOT_FOUND) {
+        // handle the not found error and check for a working-copy
```
nvisal1: Why did you decide to handle this error from within the gateway? I am not saying that it is wrong, just curious.

nrwinner: I handled it here because, as far as the caller is concerned, whether the Learning Object is a Released copy or a Working copy is unimportant. I suppose I could have handled it farther up in the Interactor, but as of right now this seemed the most logical way to compartmentalize it.

PR #463, opened 2019-08-30
Add URI's to Learning Objects

nrwinner:
This PR introduces a `resourceUris` property on Learning Objects returned from the `/learning-objects/:authorUsername/:learningObjectName` route. It can contain an object where each key corresponds to a property on the Learning Object and the value for each key is a URI for fetching that property. 

This will enable to details page (and other areas of the application as this pattern permeates through the system) to handle outages more gracefully, load different pieces of a single view asynchronously, and stop over-fetching data.

**Sample Response**
Screen Shot 2019-08-30 at 1 14 56 PM

### DIFF
--- a/src/LearningObjects/LearningObjectInteractor.ts
+++ b/src/LearningObjects/LearningObjectInteractor.ts
@@ -259,9 +259,9 @@ export async function getLearningObjectByName({
 
     const resourceUriHost = process.env.GATEWAY_API;
 
-    learningObject.resourceUris.children = `${resourceUriHost}/learning-objects/${learningObject.id}/children/summmary`;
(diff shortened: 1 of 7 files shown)


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1139,6 +1144,8 @@ export async function getLearningObjectSummaryById({
       released: loadingReleased,
     });
 
+    learningObject.attachResourceUris(process.env.GATEWAY_API);
```
nvisal1: Can we assign this environment variable to a constant at the top of the file for ease of updating?

PR #469, opened 2019-09-09
Remove zero byte file check

nrwinner:
This PR removes the zero-byte file check since it throws errors after the uploads complete and is not something that is necessarily indicative of a corrupted file.


nvisal1:
🤢

PR #496, opened 2019-11-22
Releasing a revision should not delete files from previous version. We should store old files in S3 as backups.

pzalep1:


src/FileMetadata/FileMetadataModule.ts
```diff
@@ -58,4 +59,5 @@ export class FileMetadataModule extends ExpressServiceModule {
   static getFileMetadata = getFileMetadata;
   static getAllFileMetadata = getAllFileMetadata;
   static deleteAllFileMetadata = deleteAllFileMetadata;
+  static deleteAllS3Files = deleteAllS3Files;
```
nvisal1: This function should be moved to the file manager module. We divided the functionality for managing file objects and file metadata into separate modules.

src/FileMetadata/interactors/deleteAllS3Files/deleteAllS3Files.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Only authors, admins, and editors can delete files
+ *
+ * @export
+ * @param {Requester} requester [Information about the requester]
+ * @param {string} learningObjectId [Id of the LearningObject the file belongs to]
+ *
+ * @returns {Promise<void>}
+ */
+export async function deleteAllS3Files({
```
nvisal1: We should come up with a different naming convention for this. The business logic should be driver-agnostic.

src/FileManager/interactors/deleteAllFiles/deleteAllFiles.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * Only authors, admins, and editors can delete files
+ *
+ * @export
+ * @param {Requester} requester [Information about the requester]
+ * @param {string} learningObjectId [Id of the LearningObject the file belongs to]
+ *
+ * @returns {Promise<void>}
+ */
+export async function deleteAllFiles({
```
nvisal1: This is picky, but it might be a good idea to change the name of this function. It sounds like it deletes every file in the S3 bucket.

src/FileManager/interactors/deleteAllFiles/deleteAllFiles.ts
```diff
@@ -0,0 +1,60 @@
...
+  requester,
+  learningObjectId,
+}: {
+  requester: Requester;
+  learningObjectId: string;
+}): Promise<void> {
+  try {
+    validateRequestParams({
```
nvisal1: I know that this authorization logic was not added in this PR, but we are not following this pattern anywhere else in our platform. It might be worth discussing. Do you like this approach or do you think it should be updated for consistency?

src/FileManager/interactors/deleteAllFiles/deleteAllFiles.ts
```diff
@@ -0,0 +1,60 @@
...
+    authorizeWriteAccess({ learningObject, requester });
+
+    Gateways.fileManager()
+      .deleteFolder({
+        authorUsername: learningObject.author.username,
+        learningObjectCUID: learningObject.cuid,
+        version: learningObject.version,
+        path: '/',
+      })
+      .catch(reportError);
```
nvisal1: I don't think we need this `.catch`. The function call is within a `try/catch`. The `handleError` function should take care of reporting.

nvisal1: You should double-check me on this. The `.catch` might be for error context.

pzalep1: From what I can tell this doesn't add any context.

src/FileManager/interactors/deleteAllFiles/deleteAllFiles.ts
pzalep1: I'm not really sure if I'm sold in either direction. I like consistency, but is there a reason that this is only used here? Is this a remnant of an older pattern?

pzalep1: @nvisal1

PR #499, opened 2019-12-17
Fix 404 for deleting a learning Object properly

pzalep1:
We were trying to fetch an object at version 0 after we have already deleted it which threw errors in the fetchInternalLearningObjectByCuid function.


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1087,17 +1087,17 @@ export async function deleteLearningObject({
     }
     await Gateways.submission().deletePreviousRelease({ learningObjectId: learningObject.id });
```
nvisal1: This line can be moved into the new conditional

PR #504, opened 2020-01-28
AMQP message implementation using rabbitMQ.

Olaniyan99:


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -668,6 +673,38 @@ export async function updateLearningObject({
       });
 
       if (releasableObject.version) {
+        // PublishAMQP message, establishes a channel to a RabbitMQ instance
+        // and sends new messages to a specified queue.
+        interface MQmessage {
```
pzalep1: Can we break this out to its own file?

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -668,6 +673,38 @@ export async function updateLearningObject({
...
+            Author: string,
+            LearningObjectName: string,
+            Version: number,
+            Cuid: string,
+          };
+          Requester: UserToken;
+        }
+        const rabbitMQ = await connect(AMQP_URL);
+        const rabbitMQChannel = await rabbitMQ.createChannel();
+        const obj: MQmessage = {
```
pzalep1: Could this be broken out into a separate function?

pzalep1: Can we name this something a little more descriptive? Maybe rabbitMessage or something similar

package.json
```diff
@@ -47,6 +49,7 @@
     "natives": "^1.1.3",
     "node-service-module": "^1.1.0",
     "pdfkit": "^0.9.0",
+    "rabbit-queue": "^4.6.0",
```
nvisal1: It looks like this PR adds two new RabbitMQ clients as dependencies. Are we using both?

src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -668,6 +673,38 @@ export async function updateLearningObject({
       });
 
       if (releasableObject.version) {
+        // PublishAMQP message, establishes a channel to a RabbitMQ instance
+        // and sends new messages to a specified queue.
+        interface MQmessage {
+          Category: number;
+          Payload: {
+            Username: string,
```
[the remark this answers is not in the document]
nvisal1: These properties should be lowercase.

src/LearningObjects/LearningObjectInteractor.ts
nvisal1: I would like to see a standard naming convention for this interface. In Notification Service it is called `ServiceEvent`. Thoughts? @Olaniyan99

src/LearningObjects/LearningObjectInteractor.ts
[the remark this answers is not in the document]
nvisal1: These properties should be lowercase.

src/LearningObjects/LearningObjectInteractor.ts
nvisal1: I would like to see a new RabbitMQ driver for encompassing this logic. We want our business logic to be implementation/driver-agnostic. I would suggest creating an interface for handling message queue events.

src/LearningObjects/LearningObjectInteractor.ts
Olaniyan99: i feel like ServiceEvent makes sense for notification service because it is receiving the message from learning object. i can try to come up something else for learning object service.

Olaniyan99: you think "ServiceProvider" works? or a standard name for both will be preferred in other to avoid confusion? @nvisal1

pzalep1: Can we call it QueueEvent or that too specific? This is the first instance of using a queue in the system.

package.json
Olaniyan99: i can only see only on my package.json file

Olaniyan99: ohh i see it, i'll fix that. i dont think they are both being used. but i'll double check that.

src/drivers/RabbitMQ/rabbitMQMessageQueue.ts
```diff
@@ -0,0 +1,33 @@
...
+    private channel: Channel;
+
+    static async buildMessageQueue() {
+        if (this.instance) {
+            return this.instance;
+        }
+        this.instance = new RabbitMQ();
+        this.instance.connection = await connect(AMQP_URL);
+        this.instance.channel = await this.instance.connection.createChannel();
+        return this.instance;
```
EVogel1999: Just a small thing here.  For the singleton pattern, I would check if the instance is undefined and then do lines 17-19 if it is, then just return the instance and the end like you have here.  This is a correct implementation just could be a tiny bit more efficient/less lines of code.

nvisal1: rabbitMMessageQueue was refactored into two classes. Take another look.

PR #505, opened 2020-02-03
Add filter for videos

pzalep1:
This PR introduces the ability to filter by the file types of materials of objects. This PR adds the following: 
* Ability to filter by video
* Appends the file types on the object on submission and on release into the the elasticsearch index
* Updates all related tests to match new learningObjectSearchDocument type


src/shared/MongoDB/HelperFunctions/getFileTypesOnObjects/getFileTypesOnObjects.ts
```diff
@@ -0,0 +1,57 @@
...
+    
+    if(learningObject.materials.urls.length > 0) {
+        learningObject.materials.urls.forEach(url => {
+            if(url.url.includes('youtu.be') || url.url.includes('youtube')) {
+                fileTypes.push('video');
+            }
+        })
+    }
+
+    files.forEach((file: FileMetadataInsert) => {
```
EVogel1999: This part could be  a switch statement.

src/shared/elasticsearch/CleanLearningObject/CleanLearningObject.spec.ts
```diff
@@ -18,29 +18,29 @@ describe('cleanLearningObject', () => {
       author: new User(stubs.user),
     } as Partial<LearningObject>);
     it('should contain the name property', () => {
-      expect(cleanLearningObjectSearchDocument(stubs.learningObject).name).toBe(stubs.learningObject.name);
+      expect(cleanLearningObjectSearchDocument(stubs.learningObject, stubs.fileTypes).name).toBe(stubs.learningObject.name);
```
nvisal1: We should keep our existing tests up-to-date. I would like to see at least one more test case that checks for fileTypes.

src/shared/elasticsearch/CleanLearningObject/CleanLearningObject.ts
```diff
@@ -1,23 +1,26 @@
...
 import { LearningObjectSearchDocument } from '../types/LearningObjectSearchDocument';
-
 /**
- * Prepares a Learning Object for indexing by removing unneeded data.
+ * Prepares a Learning Object for indexing by removing unneeded data and adding the fileTypes on the Learning Object
  *
  * @param learningObject A fully qualified Learning Object where all values
  * are loaded to their correct values (no foreign key IDs)
  */
 export function cleanLearningObjectSearchDocument(
```
nvisal1: Maybe the wrapper can be named `newLearningObjectSearchDocument` and the private function can be named `generateLearningObjectSearchDocument`. Thoughts?

src/shared/elasticsearch/FormatUpdateQueryParam/FormatUpdateQueryParam.spec.ts
```diff
@@ -36,14 +36,15 @@ const learningObjectSearchDocumentStub: LearningObjectSearchDocument = {
...
+        `ctx._source.author.name = \"author_name\";ctx._source.author.username = \"author_username\";ctx._source.author.email = \"author_email\";ctx._source.author.organization = \"author_organization\";ctx._source.collection = \"collection\";ctx._source.contributors[0].name = \"contributor_name\";ctx._source.contributors[0].username = \"contributor_username\";ctx._source.contributors[0].email = \"contributor_email\";ctx._source.contributors[0].organization = \"contributor_organization\";ctx._source.date = \"date\";ctx._source.description = \"description\";ctx._source.cuid = \"46e3d323-a731-4435-93a0-684c1ccfad34\";ctx._source.id = \"id\";ctx._source.length = \"length\";ctx._source.levels.0 = \"level\";ctx._source.name = \"name\";ctx._source.outcomes[0]._id = \"outcome_id\";ctx._source.outcomes[0]._bloom = \"remember and understand\";ctx._source.outcomes[0]._verb = \"remember\";ctx._source.outcomes[0]._text = \"outcome_text\";ctx._source.outcomes[0]._mappings = \"\";ctx._source.version = \"0\";ctx._source.status = \"unreleased\";ctx._source.fileTypes.0 = \"video\";ctx._source.fileTypes.1 = \"powerpoint\";ctx._source.fileTypes.2 = \"document\";`;
```
nvisal1: Doesn't the Elasticsearch package include an easier solution for formatting this?

pzalep1: We can drop in a parameter object if we would like. The tests will also have be updated to take the updated type for the formattedString. I can check to see how this trickles to other functions before I start changing anything.

src/shared/elasticsearch/CleanLearningObject/CleanLearningObject.ts
pzalep1: I think newLearningObjectSearchDocument would be good for this case with generateLearningObjectSearchDocument. Now that we're appending new bits of data that aren't on the learning object we should probably be more explicit in saying this isn't just a learning object that is being formatted for elastic, but has it's own propreties.

src/Revisions/ElasticsearchDriver.ts
```diff
@@ -21,11 +22,12 @@ export class ElasticsearchDriver implements RevisionsSearchIndex {
...
 
 async insertLearningObject(learningObject: LearningObject): Promise<void> {
-    try {
+  const fileTypes = await getFileTypesOnObjects(learningObject);
+  try {
       await this.client.index({
         index: INDEX_NAME,
         type: '_doc',
-        body: cleanLearningObjectSearchDocument(learningObject),
+        body: cleanLearningObjectSearchDocument(learningObject, fileTypes),
```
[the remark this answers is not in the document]
pzalep1: See comment below.

src/shared/MongoDB/HelperFunctions/getFileTypesOnObjects/getFileTypesOnObjects.ts
```diff
@@ -0,0 +1,55 @@
...
+export async function getFileTypesOnObjects(
+    learningObject: LearningObject,
+) {
+    let fileTypes: string[] = [];
+    const db = MongoConnector.client().db('file-service');
+
+    const files = await db.collection('files').find({ learningObjectId: learningObject.id }).toArray();
+
+    if (learningObject.materials.urls.length > 0) {
+        learningObject.materials.urls.forEach(url => {
```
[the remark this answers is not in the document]
pzalep1: I renamed this to link. I think it follows the rest of our system because a link has a title and then the url.

src/shared/elasticsearch/FormatUpdateQueryParam/FormatUpdateQueryParam.spec.ts
pzalep1: If we were to change the way that we format this we would need to write a new function that formats an object. These tests would also have to be changed to comply with the new function and its return values. I can put this in clubhouse as a chore, but I think that this is out of scope of this PR.

PR #507, opened 2020-02-10
Save downloads to its own collection

pzalep1:
This PR saves each download to its own collection. This PR needs to be merged before https://github.com/Cyber4All/library-service/pull/158.


src/shared/MongoDB/HelperFunctions/updateDownloads/updateDownloads.ts
```diff
@@ -0,0 +1,35 @@
...
+    await db.collection(COLLECTIONS.DOWNLOADS).insert(downloadObject);
+}
+
+
+function cleanObject(user: UserToken, object: LearningObject) {
+    const downloadObject = {
+        learningObjectCuid: object.cuid,
+        learningObjectVersion: object.version,
+        timestamp: Date.now(),
+        downloadedBy : {
```
EVogel1999: See bitbucket comments.

src/shared/MongoDB/HelperFunctions/updateDownloads/updateDownloads.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @param username The username of the user downloading the object
+ * @param id The learningObject ID of the learningObject being downloaded
+ */
+export async function updateDownloads(
+    user: UserToken,
+    downloadedLearningObject: LearningObject,
+): Promise<void> {
+    const downloadObject = cleanObject(user, downloadedLearningObject);
+    const db = MongoConnector.client().db('cart-service');
+    await db.collection(COLLECTIONS.DOWNLOADS).insert(downloadObject);
```
nvisal1: I like that we made a new collection.
What was your thought process behind placing the new collection in the Library database?

pzalep1: The collection already existed in Library, but I didn't change it because I wanted to keep all of the metrics in one database. If we were to start breaking our services out and implementing them separately saves and downloads are the two metrics that we use to gauge the use or "popularity" of a learning object and they should remain in the same database because they all belong to library service at the end of the day. The same way that a real library keeps track of their most popular borrows and puts them front and center when people walk in.

src/drivers/UtilityDriver.ts
```diff
@@ -0,0 +1,25 @@
...
+      uri: '',
+      json: true,
+      headers: {
+        Authorization: `Bearer ${generateServiceToken()}`,
+      },
+      method: 'GET',
+      body: {},
+    };
+
+    public async getUtilityUsers(): Promise<any[]> {
```
nvisal1: We should create a schema for Utility users instead of referencing `any`.

src/drivers/UtilityDriver.ts
```diff
@@ -0,0 +1,25 @@
...
+      try {
+        const options = { ...this.options };
+        options.uri = UTILITY_SERVICE_ROUTES.USERS();
+        return request(options);
+      } catch (e) {
+        return Promise.reject(`Unexpected error has occured while trying to get Utility Users ${e}`);
```
nvisal1: We should throw an error here instead of rejecting a string

src/shared/MongoDB/HelperFunctions/updateDownloads/updateDownloads.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Appends the downloaded boolean to the document in the users library on download
+ * @param username The username of the user downloading the object
+ * @param id The learningObject ID of the learningObject being downloaded
+ */
+export async function updateDownloads(
+    user: UserToken,
+    downloadedLearningObject: LearningObject,
+): Promise<void> {
+    const downloadObject = cleanObject(user, downloadedLearningObject);
+    const db = MongoConnector.client().db('cart-service');
```
EVogel1999: Do you think the DB name should be a constant too?  I can't remember if we have constants for DBs or if it is only collections.

pzalep1: I'm not sure if we do here. I will check

pzalep1: We only have constants for collections not databases. Realistically we wouldn't want learning object service reaching into cart, but we share datastores. This is also the only place that we are calling this database in the entire service.

PR #508, opened 2020-02-11
Create new route to get cuid

EVogel1999:
Creates a new route in learning object service that returns the cuid of a learning object given the object's name and author username.


src/drivers/express/ExpressRouteDriver.ts
```diff
@@ -42,6 +42,20 @@ export class ExpressRouteDriver {
       });
     });
 
+    router.get('/learning-objects/:username/:learningObjectName', async (req, res) => {
```
nvisal1: What was your thought process when structuring this route?

nvisal1: I would like to see functional tests for this route. We need to have functional tests in order to validate that 
- the route contract is being enforced
- user input cannot lead to a 500 series error

I think that this PR is a great opportunity to introduce it to Learning Object service. I would recommend creating swagger documentation for this single route (not the entire service). You can then feed the swagger.json into a package of your choice to generate an API client that can be used for functional testing. Functional testing is synonymous with integration testing. It simply means that the route should be tested without stubs.

Newman (https://www.npmjs.com/package/newman) seems like an awesome tool for this.

src/interactors/LearningObjectInteractor.ts
```diff
@@ -79,6 +79,36 @@ export class LearningObjectInteractor {
     return authorId;
   }
 
+  /**
+   * Finds a learning object CUID given the learning object name and
+   * the author username.
```
nvisal1: This comment should be more specific. What errors does this function throw? When is that error thrown?

The name of this function seems verbose. Is there another way of obtaining a Learning Object's CUID? 
If not, the function can be called getLearningObjectCUID. `ByAuthorAndName` will lead developers to believe that there are multiple ways to perform the same action in the service.

src/drivers/express/ExpressRouteDriver.ts
EVogel1999: The only thought process with this route structuring is that it was a mirror of what was in clark gateway at the time.  For some reason there was a clark gateway route but not a corresponding learning object route.  Up for changing this route structure though since this was legacy.

EVogel1999: > I would like to see functional tests for this route. We need to have functional tests in order to validate that
> 
> * the route contract is being enforced
> * user input cannot lead to a 500 series error
> 
> I think that this PR is a great opportunity to introduce it to Learning Object service. I would recommend creating swagger documentation for this single route (not the entire service). You can then feed the swagger.json into a package of your choice to generate an API client that can be used for functional testing. Functional testing is synonymous with integration testing. It simply means that the route should be tested without stubs.
> 
> Newman (https://www.npmjs.com/package/newman) seems like an awesome tool for this.

From what I have been reading, there are multiple ways to implement this besides using an external package.  Postman supports testing whole route collection suites, imported from the swagger.json file, and the UI package that I am using to generate swagger docs allow you to test them in the browser.  Do you think this enough for functional testing or were you thinking more along the lines of integrating it into the CI/CD pipeline?

src/interactors/LearningObjectInteractor.ts
```diff
@@ -79,6 +79,45 @@ export class LearningObjectInteractor {
...
+   */
+  public static async getLearningObjectCUID(params: {
+    dataStore: DataStore;
+    username: string;
+    learningObjectName: string;
+  }): Promise<string> {
+    const { dataStore, username, learningObjectName } = params;
+    if (!username || !learningObjectName) {
+      throw new ResourceError(
+        `No learning object name or author username specified in request`,
```
pzalep1:
```suggestion
        `Request does not contain a Learning Object name or author username.`,
```

src/interactors/LearningObjectInteractor.ts
```diff
@@ -79,6 +79,45 @@ export class LearningObjectInteractor {
...
+        ResourceErrorReason.BAD_REQUEST
+      );
+    }
+    const authID = await this.findAuthorIdByUsername({ dataStore, username: username });
+    const objectID = await dataStore.findLearningObjectByName({authorId: authID, name: learningObjectName});
+    if (!objectID) {
+      throw new ResourceError(
+        `No learning object with author username ${username} and learning object name ${learningObjectName} exists`,
```
pzalep1:
```suggestion
        `No Learning Object with author username ${username} and Learning Object name ${learningObjectName} exists`,
```

EVogel1999:
If you globally install newman, you can now run the command "newman run Learning Object Service.postman_collection.json."  To run tests and get results.  You have to put the swagger.json file into postman and then export the collection from there to get the 'Learning Object Service.postman_collection.json' file.

src/drivers/express/ExpressRouteDriver.ts
```diff
@@ -35,13 +35,100 @@ export class ExpressRouteDriver {
...
+     *  description: Public learning object API routes
+     */
+
+    /**
+     * @swagger
+     * path:
+     *  /:
+     *    get:
+     *      summary: Get learning object API welcome
+     *      tags: [Public]
```
nvisal1: Why is this commented out?

.circleci/config.yml
```diff
@@ -26,7 +35,22 @@ jobs:
...
+
+  newman-collection-run:
+    <<: *defaults
+    docker:
+      - image: circleci/node:8
+    executor: newman/postman-newman-docker
+    steps:
+      - checkout
+      - run:
+          name: Install Newman
```
nvisal1: You shouldn't need to install Newman globally if you are using the official Orb

PR #515, opened 2020-03-18
Fix duplicate es document issue on revision release

nvisal1:
***Purpose***
This PR addresses Clubhouse story https://app.clubhouse.io/clarkcan/story/1973/investigate-issue-with-revision-release


src/LearningObjects/LearningObjectInteractor.ts
```diff
@@ -1098,6 +1098,8 @@ export async function deleteLearningObject({
       );
     }
     authorizeWriteAccess({ learningObject, requester });
+
```
pzalep1: Can we add a comment so we don't forget why we moved it out?

PR #517, opened 2020-03-26
Stop from sending 500 for unauthenticated 

pzalep1:
If the requester is undefined this route currently throws a 500 because the user can't be authenticated for unreleased learning objects. The details just spins and does not handle the error correctly.


src/LearningObjects/adapters/LearningObjectRouteHandler.ts
```diff
@@ -92,7 +92,13 @@ export function initializePublic({
       }
 
       const results = await LearningObjectInteractor.getInternalLearningObjectByCuid({ dataStore, requester, authorUsername, cuid, version: version >= 0 ? version : undefined });
-
+      if (results.length > 1) {
+        if (results[0].version < results[1].version) {
```
EVogel1999: Is there a case where both versions could be equal?

PR #518, opened 2020-07-08
Add delete outcomes call

pzalep1:
This PR adds a call to Outcome-service to delete the learning outcomes of a learning object when an object is deleted.

### DIFF
--- a/src/drivers/OutcomeDriver.ts
+++ b/src/drivers/OutcomeDriver.ts
@@ -1,0 +1,26 @@
+import { OUTCOME_SERVICE_ROUTES } from '../shared/routes';
+import * as request from 'request-promise';
+import { generateServiceToken } from './TokenManager';
+import { ResourceError, ResourceErrorReason} from '../shared/errors';
+
+export class OutcomeDriver {
+    private options = {
+        uri: '',
+        json: true,
+        headers: {
+            Authorization: `Bearer ${generateServiceToken()}`,
+        },
+        method: 'DELETE',
+        body: {},
+    };
+
+    public async deleteAllOutcomes(id: string, username: string): Promise<void> {
+        try {
+            const options = { ...this.options };
+            options.uri = OUTCOME_SERVICE_ROUTES.DELETE_ALL(id, username);
+            return request(options);
+        } catch (e) {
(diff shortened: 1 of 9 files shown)


src/drivers/OutcomeDriver.ts
```diff
@@ -0,0 +1,26 @@
+import { OUTCOME_SERVICE_ROUTES } from '../shared/routes';
+import * as request from 'request-promise';
+import { generateServiceToken } from './TokenManager';
+import { ResourceError, ResourceErrorReason} from '../shared/errors';
+
+export class OutcomeDriver {
```
davidnso, line 6: a type definition like ExternalServiceRequest would be really useful in all our services since I've seen request-promise used all over in these cases 
```
uri: string ,
json: boolean, 
headers: { Authorization: string }, 
method: string,
body: {[x:sting]:string} ```

Comment (pzalep1) [reply src/drivers/OutcomeDriver.ts:6]: I'll take a look in the other services and see if we have a type somewhere and then I'll take a look at implementing this.

File: src/shared/interfaces/DataStore.ts [hunk 2]
```diff
@@ -202,8 +202,8 @@ export interface DataStore
    */
 
   // Learning Objects
-  deleteLearningObject(id: string): Promise<void>;
-  deleteMultipleLearningObjects(ids: string[]): Promise<void>;
+  deleteLearningObject(id: string, username: string): Promise<void>;
+  deleteMultipleLearningObjects(ids: any[]): Promise<void>;
```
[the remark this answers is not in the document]
pzalep1: Yes. I'll add that.

src/drivers/OutcomeDriver.ts
pzalep1: I did a little research and it appears that the package we have been using for this functionality was deprecated in February because it depends on the request package which has been deprecated as well. I am looking into another option for packages to use to make service requests.

pzalep1: This is what is used in Outcome Service: https://visionmedia.github.io/superagent/

pzalep1: I updated this to use the superagent package. I looked into a few others, but I think this package is the easiest to use.

PR #521, opened 2020-08-24
Collection Metrics: Add or extend existing route in learning object service to retrieve statistics for a specific collection

Olaniyan99:
Purpose of this PR is to get metrics for a specific collection. My approach here was to create a route that reaches out to library service where logic is stored and return them.


src/LearningObjectStats/LearningObjectStatsRouteHandler.ts
```diff
@@ -28,6 +28,24 @@ export function initialize({
...
     }
   };
+/**
+ * Initializes an express router with endpoints to get stats for a specific collection
+ */
+  const getCollectionStats = async (req: Request, res: Response) => {
+    try {
+
+      const collection = req.params.collection;
+      const metrics = await LearningObjectStatsInteractor.getCollectionMetric(collection, dataStore);
```
pzalep1, line 38: We should create a type for this.

src/drivers/MongoDriver.ts
```diff
@@ -1198,6 +1198,19 @@ export class MongoDriver implements DataStore {
...
   }
 
+  async fetchCuidInCollection(collection: string): Promise<any[]> {
+    const learningObjectCuid = await this.db.collection(COLLECTIONS.LEARNING_OBJECTS)
+    .find({$and: [{collection: collection}, {status: 'released'}]}).project({cuid: 1, _id:0}).toArray();
+
+    return learningObjectCuid;
+  }
+
+  async fetchObjectInReview(collection: string): Promise<any[]> {
```
pzalep1, line 1208: fetchCollectionObjectsInReview

src/drivers/MongoDriver.ts
```diff
@@ -1198,6 +1198,19 @@ export class MongoDriver implements DataStore {
     }
   }
 
+  async fetchCuidInCollection(collection: string): Promise<any[]> {
```
pzalep1, line 1201: fetchCollectionObjects

src/drivers/MongoDriver.ts
```diff
@@ -1198,6 +1198,19 @@ export class MongoDriver implements DataStore {
...
+    const learningObjectCuid = await this.db.collection(COLLECTIONS.LEARNING_OBJECTS)
```
pzalep1, line 1202: learningObjectCuids

src/drivers/MongoDriver.ts
```diff
@@ -1198,6 +1198,19 @@ export class MongoDriver implements DataStore {
...
+    .find({$and: [{collection: collection}, {status: 'released'}]}).project({cuid: 1, _id:0}).toArray();
+
+    return learningObjectCuid;
+  }
+
+  async fetchObjectInReview(collection: string): Promise<any[]> {
+    const obj = await this.db.collection(COLLECTIONS.LEARNING_OBJECTS)
+    .find({$and: [{collection: collection}, {status: 'review'}]}).toArray();
```
pzalep1, line 1210: This should include any waiting or proofing objects as well

package.json
```diff
@@ -48,6 +48,7 @@
     "multer": "^1.3.0",
     "natives": "^1.1.3",
     "node-service-module": "^1.1.0",
+    "nodemon": "^2.0.4",
```
EVogel1999, line 51: Small issue but shouldn't nodemon be a dev dependency?  We don't use nodemon in production do we?

Olaniyan99: u are right, i checked and we currently have nodemon runnig in prod and we dont need it

src/shared/entity/learning-object/learning-object.ts
```diff
@@ -706,6 +706,21 @@ export namespace LearningObject {
...
+  export interface CollectionMetrics {
+    authorCount:number;
+    saves: number;
+    downloads: number;
+    statusMetric: {
+      released: number
+      inReview: number;
+      waiting: number;
+      proofing: number;
+    }[];
```
pzalep1, line 718: Why does this need to be an array?

Olaniyan99: im using mongo aggregation which is returning an array of the status count.

pzalep1: From my reading of the aggregation, the array will always have a length of 1 correct? If we can we should return an object instead of an array of objects.

src/LearningObjectStats/LearningObjectStatsInteractor.ts
```diff
@@ -35,3 +38,23 @@ export async function getStats(params: {
...
+    const collectionAuthors = await dataStore.fetchAuthorIDs(collection);
+    let authors: any[] = [];
+    collectionAuthors.map(async el => {
+      const contributors: any [] = el.contributors;
+      contributors.map(e => {if (authors.includes(e) === false) { authors.push(e)}});
+      if (authors.includes(el.authorID) === false) { authors.push(el.authorID)}
+    });
+    const getInstance = new LibraryDriver();
+    // make request to library service for collection stats.
+    const getMetrics = getInstance.getCollectionMetrics(collection, statusMetrics, authors);
```
EVogel1999, line 55: This needs an await since getCollectionMetrics is async.

pzalep1 (approved):

EVogel1999 (approved):
Looks good!

